PR #8, opened 2018-04-06
Improvements

kirmorozov:
Getting real image of product at all times. 
Performance improvement: altering DOM while html parsing is in progress is bad practice. 
Removing strict version, version is simply git tag in github.
Added source as optional parameter for subscription.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,6 +2,7 @@
     "name": "klaviyo/magento2-extension",
     "description": "Klaviyo extension for Magento 2. Allows pushing newsletters to Klaviyo's platform and more.",
     "type": "magento2-module",
+    "version": "1.0.4",
     "autoload": {
         "files": [
             "registration.php"
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,6 @@
     "name": "klaviyo/magento2-extension",
     "description": "Klaviyo extension for Magento 2. Allows pushing newsletters to Klaviyo's platform and more.",
     "type": "magento2-module",
-    "version": "1.0.3",
     "autoload": {
         "files": [
             "registration.php"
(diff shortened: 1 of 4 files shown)


view/frontend/templates/analytics/initialize.phtml
```diff
@@ -2,15 +2,23 @@
     <script text="text/javascript">
         var _learnq = _learnq || [];
         _learnq.push(['account', '<?php echo $this->getPublicApiKey(); ?>']);
-        (function () { var b = document.createElement('script'); b.type = 'text/javascript'; b.async = true; b.src = ('https:' == document.location.protocol ? 'https://' : 'http://') + 'a.klaviyo.com/media/js/analytics/analytics.js'; var a = document.getElementsByTagName('script')[0]; a.parentNode.insertBefore(b, a); })();
+        window.addEventListener('load', function(event) {
+            (function () {
+                var b = document.createElement('script');
+                b.type = 'text/javascript';
+                b.async = true;
+                b.src = ('https:' == document.location.protocol ? 'https://' : 'http://') + 'a.klaviyo.com/media/js/analytics/analytics.js';
```
kirmorozov: It is possible to include http at all times, it will be fine too.

henrylamchan: Probably should leave it as is in the event that someone decides to not use https for some strange reason

composer.json
```diff
@@ -2,7 +2,6 @@
     "name": "klaviyo/magento2-extension",
     "description": "Klaviyo extension for Magento 2. Allows pushing newsletters to Klaviyo's platform and more.",
     "type": "magento2-module",
-    "version": "1.0.3",
```
henrylamchan: Totally with you that we don't need this for the case that we're dumping this on packagist and distributing through composer but I think we may still need this for submitting the extension to the Magento Marketplace. Could you add it back and also increment the minor version?

kirmorozov: Updated version.

view/frontend/templates/analytics/initialize.phtml
kirmorozov: You are allowed to connect to https at all times, but provides latency for security handshakes.

PR #11, opened 2018-05-04
fix UTM tracking after cart reclaim

robinj:
Currently UTM params are lost in the process of redirecting which makes it hard to see who has gone through the abandoned cart flow in third-party analytics solutions. We also lose the _ke param which stops Klaviyo from tracking the post-reclaim customer journey if they are reclaimed from a different device. 

This simple fix causes any URL params to be passed into the redirected URL.

### DIFF
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -41,7 +41,7 @@ class Cart extends \Magento\Framework\App\Action\Action
         }
 
         $redirect = $this->resultRedirectFactory->create();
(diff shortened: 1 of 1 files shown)


robinj:
This is ready for Review

robinj:
Closed due to lack of interest

robinj:
Reopening this for anyone who needs it. I do not use Klaviyo products anymore so I will not be able to do any further testing but it worked when I used to work at @granacom

cykolln:
This is fixed in newer releases

Issue #13, opened 2018-08-24
Getting product Image


srenon:
> [view/frontend/templates/product/viewed.phtml ](https://github.com/klaviyo/magento2-klaviyo/blob/master/view/frontend/templates/product/viewed.phtml)

Is there a reason you are getting a random image from the product image gallery using

```
$_product_image_url = null;
// Check to see if we have an image for this product.
 foreach ($_product->getMediaGalleryImages() as $_product_image) {
       if (!$_product_image->getDisabled()) {
           $_product_image_url = $_imagehelper
               ->init($_product, 'product_base_image')
               ->setImageFile($_product_image->getFile())
               ->getUrl();
                break;
       }
 }
```

instead of getting the main product image from the product?

```
$_product_image_url = $_imagehelper->init($_product, 'product_base_image')->getUrl();
```

srenon:
@remstone7 ... this is not fixed 

https://github.com/klaviyo/magento2-klaviyo/blob/master/Block/Catalog/Product/ViewedProduct.php#L155

remstone7:
will reopen, we can handle this differently and next release -> we've always grabbed the first image from the mediaGallery, certainly a better way to handle.

srenon:
Give me 10 minutes to create a PR with a fix

PR #14, opened 2018-08-24
Code refactoring / load default product image

srenon:
See issues #13


srenon:
@remstone7 .. could you take a look at and merge it into master?

srenon:
module.xml should be in the etc/module.xml (not the root folder)

see https://devdocs.magento.com/videos/fundamentals/create-a-new-module/

remstone7:
I just need to test the PR then we can merge, will do this week

PR #17, opened 2018-10-22
Fixes problem with NoSuchEntityException

JosephMaxwell:
When an email was entered (such as in a footer's newsletter subscribe box) without a customer entry in the system, and exception was thrown.


cykolln:
This was fixed with #118

PR #20, opened 2019-02-28
Fix Unexpected token error for a price >= 1000.

sippsolutions:
A price greater than 1000 will trigger an "Unexpected token" error.

Example:
`   Price: 1,337.00,`


remstone7:
it's merged from a dif contribution, thanks for the pr!

PR #23, opened 2019-04-17
add new api's

remstone7:
This PR does the following:
Version 1.0.6

- [x] get's  the current extension version `reclaim()`

- [x] returns product by item_id and quote_id, returns images

- [x] returns products by id range

- [x] Adds klaviyo.js

- [x] Add server side track ability

TODO 

- [x] Fix product id return on product-inspector

- [ ] Add auth to api endpoints

product endpoint
filter 1 is quote_id and filter_2 is item_id
image

Version endpoint
image

product-inspector endpoint:
filter_1 start id filter_2 end id
image

### DIFF
--- a/Model/Reclaim.php
+++ b/Model/Reclaim.php
@@ -70,9 +70,11 @@ class Reclaim implements ReclaimInterface
 
         return $hydrated_stores;
     }
-    public function product($filter) {
-        $quote_id = $filter['1'];
-        $item_id = $filter['2'];
+    public function product($quote_id, $item_id) {
+
+        if (!$quote_id || !$item_id){
+            throw new NotFoundException(__('quote id or item id not found'));
+        }
 
         $quote = $this->quoteFactory->create()->load($quote_id);
         if (!$quote){
@@ -97,18 +99,16 @@ class Reclaim implements ReclaimInterface
     }
 
     // handle inspector tasks to return products by id
-    public function productinspector($filter){
-        $start = $filter['1'];
-        $end = $filter['2'];
+    public function productinspector($start_id, $end_id){
 
-        if (($end - $start) > 100){
+        if (($end_id - $start_id) > 100){
             throw new NotFoundException(__('100 is the max batch'));
-        } elseif (!$start || !$end) {
+        } elseif (!$start_id || !$end_id) {
             throw new NotFoundException(__('provide a start and end filter'));
         }
 
         $response = array();
-        foreach (range($start, $end) as $number) {
+        foreach (range($start_id, $end_id) as $number) {
             $product = $this->_objectManager
                 ->create('Magento\Catalog\Model\Product')
                 ->load($number);
--- a/Model/Reclaim.php
+++ b/Model/Reclaim.php
@@ -2,8 +2,39 @@
 namespace Klaviyo\Reclaim\Model;
 use Klaviyo\Reclaim\Api\ReclaimInterface;
 
+
 class Reclaim implements ReclaimInterface
 {
+    /**
+     * Object Manager instance
+     *
+     * @var \Magento\Framework\ObjectManagerInterface
+     */
+    protected $_objectManager = null;
+    public $response;
+
+    public function __construct(
+        \Magento\Framework\ObjectManagerInterface $objectManager, 
+        \Magento\Quote\Model\QuoteFactory $quoteFactory,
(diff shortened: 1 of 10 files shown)


view/frontend/templates/analytics/initialize.phtml
```diff
@@ -1,24 +1,14 @@
 <?php if ( $this->isKlaviyoEnabled() && $this->getPublicApiKey() ): ?>
+<script type="application/javascript" async 
+ src="https://static.klaviyo.com/onsite/js/klaviyo.js?company_id=<?php echo $this->getPublicApiKey(); ?>"></script>
```
dmitrymrn: It looks like `getPublicApiKey()` (and potentially other methods) perform DB requests. It means every page load introduces an extra load. As far as I remember, Magento has cache and it should be possible to cache these values.

Model/Reclaim.php
```diff
@@ -38,4 +69,61 @@ public function stores()
...
     }
+    public function product($filter) {
+        $quote_id = $filter['1'];
+        $item_id = $filter['2'];
+        $quote = $this->quoteFactory->create()->load($quote_id);
+        
+        if (!$quote){
+            return array('error' => 'not found');
+        }
+        $item = $quote->getItemById($item_id);
```
klaviyojad: maybe throw an error as well if that item_id wasnt invalid

Model/Reclaim.php
```diff
@@ -38,4 +69,61 @@ public function stores()
...
+        return $response;
+
+    }
+    public function _getImages($product){
+        $images = $product->getMediaGalleryImages();
+        $image_array = array();
+        foreach($images as $image) {
+            $image_url = $image->getUrl();
+            if ($image_url){
+                array_push($image_array, $image_url);
```
dmitrymrn: or you could do (totally optional)
```suggestion
                $image_array[] = $image_url;
```

Model/Reclaim.php
```diff
@@ -38,4 +69,61 @@ public function stores()
...
+            if (!$product){
+                continue;
+            }
+            $data = array(
+                'id' => $product->getProductId(),
+                'sku' => $product->getSku(),
+                'type_id' => $product->getTypeId(),
+                'price' => $product->getPrice(),
+            );
+            array_push($response, $data);
```
dmitrymrn: Unlike Python, PHP passes by value and not by reference. This means that every new variable takes extra memory. While it's good to be explicit in the Python world, in PHP it comes with a cost. Here is a more efficient pattern:
```suggestion
            $response[] = array(
                'id' => $product->getProductId(),
                'sku' => $product->getSku(),
                'type_id' => $product->getTypeId(),
                'price' => $product->getPrice()
            )
```

Model/Reclaim.php
```diff
@@ -38,4 +69,61 @@ public function stores()
 
         return $hydrated_stores;
     }
+    public function product($filter) {
+        $quote_id = $filter['1'];
+        $item_id = $filter['2'];
```
dmitrymrn: Dumb question, but why Magento makes indexes to be strings?..

Model/Reclaim.php
```diff
@@ -38,4 +69,61 @@ public function stores()
...
+        
+        if (!$quote){
+            return array('error' => 'not found');
+        }
+        $item = $quote->getItemById($item_id);
+        $product = $this->_objectManager->get('Magento\Catalog\Model\Product')->load($item->getProductId());
+
+        $image_array = $this->_getImages($product);
+        
+        $response['$body'] = [
```
dmitrymrn: is `'$body'` a special Klaviyo var?

remstone7: removed, no reason for it

Model/Reclaim.php
remstone7: it's a filter from the url..

http://magento2.local:8888/rest/V1/klaviyo/reclaim/product?filter[1]=4&filter[2]=15

it takes the queryparams and filter['1'] is grabbing the key of the filter

dmitrymrn: image

Jaeto: Do we have control over the query param keys? Having to tie to `filter[1]` and `filter[2]` is rough.

remstone7: nice find, updated

PR #25, opened 2019-04-25
subscriber api endpoint

remstone7:
It appears we don't get customers who subscriber as a guest as the hooks don't fire.  this is to expose 2 api endpoints to be able to query this data.  

1. for historical sync by id's (batching)
2. for periodic date pull (since param)

Endpoints:

```
@param mixed $start
@param mixed $until
@param mixed $store_id
```
/V1/klaviyo/reclaim/customers/subscriptions/periodic
Ex
```
rest/V1/klaviyo/reclaim/customers/subscriptions/periodic?start=2019-04-25%20:00:00&until=2019-04-25%2018:00:00:00
```

```
@param mixed $start_id	   
@param mixed $end_id	     
@param mixed $store_id
```

/V1/klaviyo/reclaim/customers/subscriptions/historical

```
http://magento2.local:8888/rest/V1/klaviyo/reclaim/customers/subscriptions/historical?start_id=1&end_id=10
```

### DIFF
--- a/Model/Reclaim.php
+++ b/Model/Reclaim.php
@@ -17,6 +17,7 @@ class Reclaim implements ReclaimInterface
     public $response;
 
     const MAX_QUERY_DAYS = 10;
+    const SUBSCRIBER_BATCH_SIZE = 500;
 
     public function __construct(
         \Magento\Framework\ObjectManagerInterface $objectManager, 
@@ -29,11 +30,9 @@ class Reclaim implements ReclaimInterface
         $this->quoteFactory = $quoteFactory;
         $this->_objectManager = $objectManager;
         $this->_subscriber= $subscriber;
-        $this->subscriberCollection = $subscriberCollection;
+        $this->_subscriberCollection = $subscriberCollection;
         $this->_klaviyoHelper = $klaviyoHelper;
-
     }
-
 
     /**
      * Returns extension version
@@ -137,6 +136,12 @@ class Reclaim implements ReclaimInterface
 
     }
 
+    public function getSubscribersCount()
+    {
+        $subscriberCount =$this->_subscriberCollection->create()->count();
+        return $subscriberCount;
+    }
+
     public function getSubscribersById($start_id, $end_id, $storeId=null)
     {
         if (!$start_id || !$end_id ){ 
@@ -147,9 +152,13 @@ class Reclaim implements ReclaimInterface
             throw new NotFoundException(__('end_id should be larger than start_id'));
         }
 
+        if (($end_id - $start_id) > self::SUBSCRIBER_BATCH_SIZE){
+            throw new NotFoundException(__('Max batch size is 500'));
+        }
+
         $storeIdFilter = $this->_storeFilter($storeId);
 
-        $subscriberCollection =$this->subscriberCollection->create()
+        $subscriberCollection =$this->_subscriberCollection->create()
             ->addFieldToFilter('subscriber_id', ['gteq' => (int)$start_id])
             ->addFieldToFilter('subscriber_id', ['lteq' => (int)$end_id])
             ->addFieldToFilter('store_id', [$storeIdFilter => $storeId]);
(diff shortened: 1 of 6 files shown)


Model/Reclaim.php
```diff
@@ -127,6 +136,84 @@ public function productinspector($start_id, $end_id){
...
+        if (!$start_id || !$end_id ){ 
+            throw new NotFoundException(__('Please provide start_id and end_id'));
+        }
+
+        $storeIdFilter = $this->_storeFilter($storeId);
+
+        $subscriberCollection =$this->subscriberCollection->create()
+            ->addFieldToFilter('subscriber_id', ['gteq' => (int)$start_id])
+            ->addFieldToFilter('subscriber_id', ['lteq' => (int)$end_id])
+            ->addFieldToFilter('store_id', [$storeIdFilter => $storeId]);
```
klaviyojad: do we still wanna do that if $storeId is null?

Model/Reclaim.php
```diff
@@ -127,6 +136,84 @@ public function productinspector($start_id, $end_id){
         return $response;
 
     }
+
+    public function historicalcustomersubscription($start_id, $end_id, $storeId=null)
+    {
+
+        if (!$start_id || !$end_id ){
```
klaviyojad: can we also check that $start_id <$end_id?

Model/Reclaim.php
```diff
@@ -127,6 +136,84 @@ public function productinspector($start_id, $end_id){
...
+        }
+
+        $storeIdFilter = $this->_storeFilter($storeId);
+
+        $subscriberCollection =$this->subscriberCollection->create()
+            ->addFieldToFilter('subscriber_id', ['gteq' => (int)$start_id])
+            ->addFieldToFilter('subscriber_id', ['lteq' => (int)$end_id])
+            ->addFieldToFilter('store_id', [$storeIdFilter => $storeId]);
+
+        $subscriberCollection =$this->subscriberCollection->create()
```
klaviyojad: why do we do that after getting the start_id to end_id range?

Model/Reclaim.php
```diff
@@ -127,6 +136,84 @@ public function productinspector($start_id, $end_id){
...
+        $response = $this->_packageSubscribers($subscriberCollection);
+
+        return $response;
+    }
+
+    public function customersubscription($start, $until, $storeId=null)
+    {
+        
+        if (!$start || !$until ){ 
+            throw new NotFoundException(__('Please provide start_id and end_id'));
```
klaviyojad: its different here

klaviyojad:
I dont get the naming convention you use, periodic vs historical?  Why not just call it getSubscribedsById and getSubscribersByDateRange?

Model/Reclaim.php
remstone7: it queries for nlike storeId = null.. was basically to save some additional logic of dropping the 3rd query

Model/Reclaim.php
remstone7: oops, removing from testing

Model/Reclaim.php
```diff
@@ -12,20 +12,29 @@ class Reclaim implements ReclaimInterface
...
         \Magento\Quote\Model\QuoteFactory $quoteFactory,
+        \Magento\Newsletter\Model\Subscriber $subscriber,
+        \Magento\Newsletter\Model\ResourceModel\Subscriber\CollectionFactory $subscriberCollection,
         \Klaviyo\Reclaim\Helper\Data $klaviyoHelper
         )
     {
         $this->quoteFactory = $quoteFactory;
         $this->_objectManager = $objectManager;
+        $this->_subscriber= $subscriber;
+        $this->subscriberCollection = $subscriberCollection;
```
remstone7: should `_` this

Model/Reclaim.php
```diff
@@ -127,6 +136,81 @@ public function productinspector($start_id, $end_id){
...
+        // $start = '2019-04-25 00:00:00';
+
+        // don't want any big queries, we limit to 10 days
+        $until_date = strtotime($until);
+        $start_date = strtotime($start);
+        $datediff = $until_date - $start_date;
+
+        if (abs(round($datediff / (60 * 60 * 24))) > self::MAX_QUERY_DAYS){
+            throw new NotFoundException(__('Please don\'t query for more than 10 days'));
+        }
```
invertedfjord: maybe change the wording on this to something like "Cannot query more than 10 days".

Model/Reclaim.php
```diff
@@ -127,6 +136,81 @@ public function productinspector($start_id, $end_id){
...
+        return $response;
+    }
+
+    public function getSubscribersByDateRange($start, $until, $storeId=null)
+    {
+        
+        if (!$start || !$until ){ 
+            throw new NotFoundException(__('Please provide start and until param'));
+        }
+        // start and until date formats
```
invertedfjord: should we validate date formats as well?

etc/webapi.xml
```diff
@@ -18,6 +18,18 @@
...
+            <resource ref="anonymous"/>
+        </resources>
+    </route>
+    <route url="/V1/klaviyo/reclaim/customers/subscriptions/historical" method="GET">
+        <service class="Klaviyo\Reclaim\Api\ReclaimInterface" method="getSubscribersById"/>
+        <resources>
+            <resource ref="anonymous"/>
+        </resources>
+    </route>
     <route url="/V1/klaviyo/reclaim/" method="GET">
```
invertedfjord: Do we want to add authentication. Doesn't seem too likely, but theoretically we could grab customers subscribers.

Model/Reclaim.php
```diff
@@ -12,20 +12,29 @@ class Reclaim implements ReclaimInterface
...
         $this->quoteFactory = $quoteFactory;
         $this->_objectManager = $objectManager;
+        $this->_subscriber= $subscriber;
+        $this->subscriberCollection = $subscriberCollection;
         $this->_klaviyoHelper = $klaviyoHelper;
 
     }
 
+
     /**
```
invertedfjord: Nit. remove extra line.

etc/webapi.xml
remstone7: yea adding for these

PR #26, opened 2019-05-15
bis api

remstone7:


Model/Reclaim.php
```diff
@@ -106,6 +111,52 @@ public function product($quote_id, $item_id) {
...
+    */
+    public function backinstock($product_id)
+    {
+        if (!$product_id){
+            throw new NotFoundException(_('A product id is required'));
+        }
+        $product = $this->_objectManager->create('Magento\Catalog\Model\Product')->load($product_id);
+
+        if (!$product){
+            throw new NotFoundException(_('A product was not found'));
```
klaviyojad: maybe call out the product_id passed in the Exception

Model/Reclaim.php
```diff
@@ -106,6 +111,52 @@ public function product($quote_id, $item_id) {
...
+            'price' => $product->getPrice(),
+            'available' => true,
+            'inventory_quantity' => $this->_stockItem->getStockQty($productId),
+            'inventory_policy' => $this->_getStockItem($productId)
+        ));
+        // check to see if the product has variants, if it doesn't just return the product information
+        try {
+            $_children = $product->getTypeInstance()->getUsedProducts($product);
+            // throws a fatal error, so catch it generically and return
+        } catch (\Error $e) {
```
klaviyojad: is that a typo?

Model/Reclaim.php
```diff
@@ -106,6 +111,52 @@ public function product($quote_id, $item_id) {
...
+        $product = $this->_objectManager->create('Magento\Catalog\Model\Product')->load($product_id);
+
+        if (!$product){
+            throw new NotFoundException(_('A product was not found'));
+        }
+
+        $productId = $product->getId();
+
+        $response = array(array(
+            'id' => $productId,
```
dmitrymrn: image


It looks like Magento's API doesn't support IDs and only supports SKUs. 
https://devdocs.magento.com/guides/v2.3/rest/list.html

Can you please change this to:
```suggestion
            'id' => $product->getSku(),
```

Model/Reclaim.php
```diff
@@ -106,6 +111,52 @@ public function product($quote_id, $item_id) {
...
+        try {
+            $_children = $product->getTypeInstance()->getUsedProducts($product);
+            // throws a fatal error, so catch it generically and return
+        } catch (\Error $e) {
+            return $response;
+        }
+        
+        foreach ($_children as $child){
+            $response['variants'][] = array(
+                'id' => $child->getId(),
```
dmitrymrn: And here too please:
```suggestion
                'id' => $child->getSku(),
```

Model/Reclaim.php
dmitrymrn: Aargh... 🤕We need a product ID, which we can use to find its SKU. Let's keep this as is. But we still need to pass variant SKU.

dmitrymrn: image

I think we also need to pass the store ID for both product and variant... 🤢

Model/Reclaim.php
dmitrymrn: This is a way to set the local scope. Without it, PHP would try to find `\Model\Reclaim\Error`

Model/Reclaim.php
remstone7: good call on store id, will make it optional arg

Model/Reclaim.php
```diff
@@ -106,6 +113,59 @@ public function product($quote_id, $item_id) {
...
+            $product = $this->_productFactory->create()->setStoreId($store_id)->load($product_id);    
+        } else {
+            $product = $this->_productFactory->create()->load($product_id);
+        }
+
+        if (!$product){
+            throw new NotFoundException(_('A product with id '. $product_id .' was not found'));
+        }
+
+        $productId = $product->getId();
```
Jaeto: If `$product_id` is provided in the signature why do you need to get it again? Is this value different?

Model/Reclaim.php
```diff
@@ -235,4 +295,9 @@ public function _getImages($product){
         }
         return $image_array;
     }
+    public function _getStockItem($productId)
+    {
+        $stock = $this->_stockItemRepository->get($productId);
+        return $stock->getManageStock();
```
Jaeto: do you need the var to return vs just retuning the `this->_stockItemRepository->($productId);`

Issue #29, opened 2019-06-25
Customer Specific code in .phtml files


srenon:
/view/frontend/templates/analytics/initialize.phtml

```
    <script text="text/javascript">
        var _learnq = _learnq || [];
        <?php if ( $this->isLoggedIn() ): ?>
            _learnq.push(['identify', {
                $email: "<?php echo $this->getCustomerEmail() ?>",
                $first_name: "<?php echo $this->getCustomerFirstname(); ?>",
                $last_name: "<?php echo $this->getCustomerLastname(); ?>"
            }]);
        <?php endif; ?>
    </script>
```

You should NOT load customer specific code in phtml files because that block will get cached by FPC and then every customer will get the same content. So after cleaning cached if the first person who visits your site logged in then the there is a strong possibility that all other users who view page source will see that user info. This could become a privacy issue and also the wrong info will be sent to you. You should instead use Javascript to set this information

remstone7:
Thanks, taking a look at this one!

srenon:
A quick fix would be to create a plugin for Magento\Customer\CustomerData\Customer

    public function afterGetSectionData(Customer $subject, $result)
    {
        if ($this->currentCustomer->getCustomerId()) {
            $result['lastname'] = $this->currentCustomer->getLastname();
            $result['email'] = $customer->getEmail();
        } else {
            $result['lastname'] = '';
            $result['email'] = '';
        }
        return $result;
    }

Then move that logic to a JS files

    define([
         'Magento_Customer/js/customer-data'
    ], function (customerData) {
         var customer = customerData.get('customer')()
         var _learnq = _learnq || [];
         if  (customer.email) {
               _learnq.push(['identify', {
                $email:  customer.email        ,
                $first_name: customer.firstname,
                $last_name: customer.lastname
            }]);
     ....

remstone7:
@srenon doing some research, it looks like a dependency injection is okay to utilize as it seems common practice? https://www.etatvasoft.com/blog/dependency-injection-in-magento-two/  & https://devdocs.magento.com/guides/v2.3/extension-dev-guide/depend-inj.html

I did some testing and doesn't appear cacheing will cause an issue as i've been able to trigger this through acting as multiple users without any issues.

srenon:
If you are using advance Cloudflare setting or any other caching service then the first person who visits the site page will get caches after clearing the cache. 

If this was the acceptable approach why did Magento use js to load the welcome message in the header?

[...]

PR #31, opened 2019-07-16
Issue #29 - Remove customer Specific info from .phtml files

srenon:


view/frontend/web/js/customer.js
```diff
@@ -0,0 +1,23 @@
...
+ * https://www.magepal.com | support@magepal.com
+ */
+
+define([
+    'underscore',
+    'Magento_Customer/js/customer-data'
+], function (_, customerData) {
+    'use strict';
+    
+    window._learnq = window._learnq || [];
```
remstone7: do we need the `.window` here or just use `var` ?

Plugin/Customer/Model/CustomerData.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * @param Customer $subject
+     * @param $result
+     * @return mixed
+     */
+    public function afterGetSectionData(Customer $subject, $result)
+    {
+        if ($this->currentCustomer->getCustomerId()) {
+            $result['lastname'] = $this->currentCustomer->getLastname();
```
remstone7: can we add `firstname`?

srenon: The first name is added by default magneto

see Magento\Customer\CustomerData\Customer

    public function getSectionData()
    {
        if (!$this->currentCustomer->getCustomerId()) {
            return [];
        }

        $customer = $this->currentCustomer->getCustomer();
        return [
            'fullname' => $this->customerViewHelper->getCustomerName($customer),
            'firstname' => $customer->getFirstname(),
            'websiteId' => $customer->getWebsiteId(),
        ];

view/frontend/web/js/customer.js
srenon: In a function, the variable will be a local variable to that function so you need "window." to get access to the global namespace of `_learnq`

Issue #33, opened 2019-07-19
Email address not validated before saving to quote


jasuten:
The automatic email address capture on the checkout page is not validated before saving it to the quote, this causes validation issues elsewhere:

Steps to reproduce:
- Add products to cart
- Proceed to checkout page
- Enter malformed email, e.g. example@broken
- Go back to cart page
- Remove a product from cart
- Error message appears "We can't remove the item."

remstone7:
taking a look at this, thanks!

remstone7:
can reproduce, will patch

remstone7:
fixing in https://github.com/klaviyo/magento2-klaviyo/pull/34

Issue #32, opened 2019-07-18
Swagger is not working


aurelienchl:
We're on Magento 2.2.6 and this plugin is causing an error for Swagger. Here's the issue when trying to access to this page:
/rest/all/schema?services=all
Data type "anyType" is not declared.

remstone7:
thanks for bringing that up.. fixing it w/ this https://github.com/klaviyo/magento2-klaviyo/pull/34 `@return mixed[]`

remstone7:
released!

PR #34, opened 2019-07-24
Media URLS & Fix Schema & fix checkout email 

remstone7:
creates:
image

If someone has a custom media url route, they can specify it here, it will then allow the products api endpoint to utilize that custom url

also, making sure it's a valid url 
image

handles issues #32 #33

### DIFF
--- a/Controller/Checkout/Email.php
+++ b/Controller/Checkout/Email.php
@@ -26,6 +26,15 @@ class Email extends \Magento\Framework\App\Action\Action
         $quote = $this->_objectManager->create('Magento\Checkout\Model\Cart')->getQuote();
 
         $customer_email = $this->getRequest()->getParam('email');
+        // add a quick email validation
+        $fp = fopen($customer_email, 'a');
+        fwrite($fp, $data);
+        if (filter_var($customer_email, FILTER_VALIDATE_EMAIL)) {
+            $fp = fopen('filtered', 'a');
+            fwrite($fp, $data);
+            return;
+        }
+
         $quote->setCustomerEmail($customer_email);
         $quote->save();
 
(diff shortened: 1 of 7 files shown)


Model/Reclaim.php
```diff
@@ -284,17 +284,24 @@ public function _storeFilter($storeId)
...
-                $image_array[] = $image_url;
-            }
+            $image_array[] = $this->handleMediaURL($image);
         }
         return $image_array;
     }
+    public function handleMediaURL($image)
+    {
+        $custom_media_url = $this->_klaviyoHelper->getCustomMediaURL();
+        if ($custom_media_url){
```
klaviyojad: is this false on blank strings?

remstone7: yes

klaviyojad: can we also check if image->getFile() is not null?

remstone7: everytime an image is uploaded, the file is stored.  it will always be there, it's part of magento core

etc/adminhtml/system.xml
```diff
@@ -21,6 +21,10 @@
                 <field id="private_api_key" translate="label" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
                     <label>Private Klaviyo API Key</label>
                 </field>
+                <field id="custom_media_url" translate="label" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
```
kenney: This should be 40?

Controller/Checkout/Email.php
```diff
@@ -26,6 +26,11 @@ public function execute()
         $quote = $this->_objectManager->create('Magento\Checkout\Model\Cart')->getQuote();
 
         $customer_email = $this->getRequest()->getParam('email');
+        // add a quick email validation
+        if (filter_var($customer_email, FILTER_VALIDATE_EMAIL)) {
```
kenney: Logic here is wrong

PR #40, opened 2019-07-30
Fixes error undefined method getLastname

matiaso:
Fixes error:
Uncaught Error: Call to undefined method Magento\Customer\Helper\Session\CurrentCustomer::getLastname()


remstone7:
thanks @matiaso

PR #43, opened 2019-07-30
use require js than vanilla js

remstone7:


view/frontend/templates/product/viewed.phtml
```diff
@@ -3,9 +3,9 @@
 ?>
 
 <script text="text/javascript">
-    document.addEventListener("DOMContentLoaded", function() {
-        _learnq.push(['track', 'Viewed Product', <?= $block->getViewedProductJson() ?>]);
-        _learnq.push(['trackViewedItem', <?= $block->getViewedItemJson() ?>]);
-    });
+  require(['domReady!'], function () {
```
smoucka: had to go read about the "!". very nice!

Plugin/Customer/Model/CustomerData.php
```diff
@@ -33,14 +33,12 @@ public function __construct(
      */
     public function afterGetSectionData(Customer $subject, $result)
```
klaviyojad: why is $subject part of the method if it is not used?

remstone7: magento standard to define customer on this overriding method https://magento.stackexchange.com/questions/134884/how-to-override-getsectiondata-method-of-magento-checkout-customerdata-cart-ph

PR #46, opened 2019-08-15
account for logged in users on checkout

remstone7:


view/frontend/web/js/view/checkout/email.js
```diff
@@ -1,40 +1,60 @@
 define([
     'uiComponent',
     'jquery',
-    'jquery/ui'
+    'jquery/ui',
+    '!domReady'
 ], function(Component) {
-    'use strict';
+    // initial
```
remstone7: will update this

view/frontend/web/js/view/checkout/email.js
```diff
@@ -1,40 +1,61 @@
 define([
     'uiComponent',
     'jquery',
-    'jquery/ui'
+    'jquery/ui',
+    '!domReady'
```
klaviyojad: why is this needed?  I don't see anything in the added code pertaining to it

view/frontend/web/js/view/checkout/email.js
```diff
@@ -1,40 +1,61 @@
...
 ], function(Component) {
     'use strict';
+    // initialize the customerData prior to returning the component
+    var _klaviyoCustomerData = window.customerData;
```
klaviyojad: what is window.customerData?

view/frontend/web/js/view/checkout/email.js
remstone7: it makes the code not load until the document is ready. https://requirejs.org/docs/api.html#pageload

view/frontend/web/js/view/checkout/email.js
remstone7: from our plugin, we put customerData on every page which contains a user's email if they're logged in

view/frontend/web/js/view/checkout/email.js
```diff
@@ -1,40 +1,61 @@
...
+            });
+        },
+        postUserEmail: function (customer_email) {
+          var path = window.location.pathname;
+          if (path.slice(-1) == '/' ) {
+              path = path.slice(0, -1);
+          }
 
-                var url = window.location.protocol + '//' + window.location.host + path.substring(0,path.lastIndexOf("/"));
+          var url = window.location.protocol + '//' + window.location.host + path.substring(0,path.lastIndexOf("/"));
```
klaviyojad: why all the slicing and substring then appending, can we just do all this manipulation as part of one line instead of 3 lines? whatever it is is only used once in that ajax call as far as I can see

remstone7: looks like it's some cleaning, i don't know if i have a recommendation to do this in 1 line as it would get even more bulky - open to suggestions though!

view/frontend/templates/product/viewed.phtml
```diff
@@ -4,8 +4,9 @@
 
 <script text="text/javascript">
   require(['domReady!'], function () {
-    _learnq.push(['track', 'Viewed Product', <?= $block->getViewedProductJson() ?>]);
-    _learnq.push(['trackViewedItem', <?= $block->getViewedItemJson() ?>]);
+    window._learnq = window._learnq || [];
```
remstone7: adding the window here as i'd rather this not break if it has a loading problem

ihor-sviziev:
@remstone7,
Would be great also send email to _learnq when guest customer fills it up on checkout. It will allow to associate customer flow on frontend (which products he/she saw) with info that later will be retrieved from API (quotes, orders)

remstone7:
@ihor-sviziev will add that to next release!

PR #48, opened 2019-08-17
Removed MagePal copyright from customer.js

clivewalkden:
Noticed this in the customer.js and thought it best be removed


remstone7:
a pr #46 is already out for review with this, going to close this one as it's going with new update

PR #51, opened 2019-10-29
pass email to learnq in checkout / add setting for double opt-in mode

njparadis:


Helper/Data.php
```diff
@@ -87,6 +91,15 @@ public function getNewsletter()
         return $this->getScopeSetting(self::NEWSLETTER);
     }
 
+    public function getOptInSetting()
+    {
+        if($this->getScopeSetting(self::USING_KLAVIYO_LIST_OPT_IN)) {
+            return SUBSCRIBE;
```
remstone7: need to return `self::API_SUBSCRIBE` ?

Helper/Data.php
```diff
@@ -87,6 +91,15 @@ public function getNewsletter()
...
+        } else {
```
remstone7: use an else if, i'd rather be explicit ?

etc/adminhtml/system.xml
```diff
@@ -39,6 +39,10 @@
                     <source_model>Klaviyo\Reclaim\Helper\ListOptions</source_model>
                     <frontend_model>Klaviyo\Reclaim\Block\System\Config\Form\Field\Newsletter</frontend_model>
                 </field>
+                <field id="using_klaviyo_list_opt_in" translate="label" type="radios" sortOrder="20" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Use Klaviyo Opt-In Setting For This List?</label>
+                    <source_model>Vendor\Extension\Model\Config\Source\Radiobtn</source_model>
```
remstone7: i think the path here is wrong? this needs to be scoped down to klaviyo repo?

Helper/Data.php
remstone7: actually nvm, sorry this is good - to use an else because it's a bool.. just do the return self::*

Helper/Data.php
```diff
@@ -87,6 +91,15 @@ public function getNewsletter()
         return $this->getScopeSetting(self::NEWSLETTER);
     }
 
+    public function getOptInSetting()
+    {
+        if($this->getScopeSetting(self::USING_KLAVIYO_LIST_OPT_IN)) {
+            return self::API_SUBSCRIBE;
+        } else if(!($this->getScopeSetting(self::USING_KLAVIYO_LIST_OPT_IN))) {
```
remstone7: just move this back to an else whenever you get the chance

Helper/Data.php
```diff
@@ -87,6 +91,15 @@ public function getNewsletter()
         return $this->getScopeSetting(self::NEWSLETTER);
     }
 
+    public function getOptInSetting()
+    {
+        if($this->getScopeSetting(self::USING_KLAVIYO_LIST_OPT_IN)) {
```
remstone7: nit: add a space after the if making it like: `if (`

Model/Config/Source/Radiobtn.php
```diff
@@ -0,0 +1,19 @@
...
+            [
+                'value' => true, 
+                'label' => __('Yes, use the Klaviyo settings for this list')
+            ], 
+            [
+                'value' => false, 
+                'label' => __('No, do not send opt-in emails from Klaviyo')
+            ],
+        ];
+	}
```
remstone7: Could you fix the indentation on this curly

PR #52, opened 2019-11-22
fix checkout tracking

njparadis:
learnq is now properly loaded on the checkout page


remstone7:
just bump the version

PR #54, opened 2020-01-08
Klaviyo user creation

njparadis:
First pass at streamlining the initial setup process for M2:
- Added a section to the extension config page to automate the REST user creation process
- Sensitive fields in the config page are now encrypted and obfuscated in the frontend


njparadis:
### Sensitive data updates
Obfuscated private API key:
image

Obfuscated password:
image

### Error/success messaging examples
User already exists:
image

Invalid email:
image

Crappy password:
image

Success:
image

Setup/InstallData.php
```diff
@@ -0,0 +1,75 @@
...
+        $resource = [
+            'Magento_Backend::all'
+        ];
+        try{
+            $this->rulesFactory->create()
+                ->setRoleId($role->getId())
+                ->setResources($resource)
+                ->saveRel();
+        } catch (\Exception $ex) {
+            $this->logger->info('RULE CREATION ISSUE: ' . $ex->getMessage());
```
klaviyojad: you mean role creation issue

Setup/InstallData.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->rulesFactory = $rulesFactory;
+        $this->logger = $logger;
+    }
+ 
+    public function install(
+        ModuleDataSetupInterface $setup, 
+        ModuleContextInterface $context
+    ) {
+        $role = $this->roleFactory->create();
+        $role->setName('Klaviyo')
```
klaviyojad: dunno if you want to make that a constant

Observer/RestApiUserObserver.php
```diff
@@ -0,0 +1,102 @@
...
+        $roleCollection = $role->getData();
+
+        $adminInfo = [
+            'role_id' => 1,
+            'username' => $apiUsername,
+            'firstname' => 'klaviyo',
+            'lastname'    => 'klaviyo',
+            'email'     => $apiEmail,
+            'password'  => $apiPassword,
+            'interface_locale' => 'en_US',
```
klaviyojad: what if the Magento 2 wasnt in the US?

njparadis: afaik this just determines the language used in the actual UI if you were to login as this user, will do more research

Setup/InstallData.php
njparadis: at this stage, technically the role has already been created, we are just now trying to create a rule and assign it to that role

Observer/RestApiUserObserver.php
```diff
@@ -0,0 +1,102 @@
...
+     * @param RoleCollectionFactory $_roleCollectionFactory
+     */
+    public function __construct(
+        \Magento\Framework\Message\ManagerInterface $messageManager,
+        \Klaviyo\Reclaim\Helper\Data $data_helper,
+        \Magento\User\Model\UserFactory $userFactory,
+        \Magento\Authorization\Model\ResourceModel\Role\CollectionFactory $roleCollectionFactory
+    ) {
+        $this->messageManager = $messageManager;
+        $this->data_helper = $data_helper;
```
remstone7: nit: could we just standardize camalCase?

Observer/RestApiUserObserver.php
```diff
@@ -0,0 +1,102 @@
...
+        $this->_roleCollectionFactory = $roleCollectionFactory;
+    }
+
+    public function execute(\Magento\Framework\Event\Observer $observer)
+    {
+        $apiUsername = $this->data_helper->getRestApiUsername();
+        $apiPassword = $this->data_helper->getRestApiPassword();
+        $apiEmail = $this->data_helper->getRestApiEmail();
+
+        //
```
remstone7: what's this?

Observer/RestApiUserObserver.php
```diff
@@ -0,0 +1,102 @@
...
+        $role = $this->_roleCollectionFactory->create();
+        $roleCollection = $role->getData();
+
+        $adminInfo = [
+            'role_id' => 1,
+            'username' => $apiUsername,
+            'firstname' => 'klaviyo',
```
remstone7: const this can call it like `self::KLAVIYO_FIRST_NAME`

Observer/RestApiUserObserver.php
```diff
@@ -0,0 +1,102 @@
...
+            'email'     => $apiEmail,
+            'password'  => $apiPassword,
+            'interface_locale' => 'en_US',
+            'is_active' => 1
+        ];
+
+        //try to get the ID of the Klaviyo role
+        try {
+            foreach ($roleCollection as $item) {
+                if ($item['role_name'] == 'Klaviyo' && $item['role_type'] == 'G') {
```
remstone7: constants for G and klaviyo, is there a reason it's capital K but from above, if so could we put another const (all string literals should be consts)

Setup/InstallData.php
remstone7: is it possible to create a klaviyo.log file?

Observer/RestApiUserObserver.php
remstone7: ahh i see this is in the install script

Setup/InstallData.php
```diff
@@ -0,0 +1,75 @@
+<?php
+namespace Klaviyo\Reclaim\Setup;
+  
+use \Magento\Framework\Setup\InstallDataInterface;
+use \Magento\Framework\Setup\ModuleDataSetupInterface;
+use \Magento\Framework\Setup\ModuleContextInterface;
+use \Magento\Authorization\Model\Acl\Role\Group as RoleGroup;
```
remstone7: why the as here?

remstone7:
assuming the deletion process would still be in the rest role section?

Setup/InstallData.php
njparadis: makes it easier to read IMO, otherwise it would just be Group

ihor-sviziev:
This is actually not an REST user, this is admin user, that has full permissions to the admin.

It should NOT be inside Stores => Configuration

Setup/InstallData.php
njparadis: opened a TP on the user story Kanban board to be included in a future release

Helper/Data.php
```diff
@@ -49,17 +56,40 @@ protected function getScopeSetting($path){
 
         if (isset($storeId)) {
             $scope = \Magento\Store\Model\ScopeInterface::SCOPE_STORE;
-            $value = $storeId;
-            return $this->_scopeConfig->getValue($path, $scope, $value);
+            $scopeCode = $storeId;
```
remstone7: why don't we just use `$storeId`?

Observer/KlaviyoUserObserver.php
```diff
@@ -0,0 +1,113 @@
+<?php
+
+namespace Klaviyo\Reclaim\Observer;
+
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Event\Observer;
+use \Magento\Authorization\Model\Acl\Role\Group as RoleGroup;
```
remstone7: should this be 
`Magento\Authorization\Model\Acl\Role\Group as RoleGroup;`

Observer/KlaviyoUserObserver.php
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Init
+     *
+     * @param MessageManager $_messageManager
+     * @param DataHelper $_dataHelper
+     * @param UserFactory $_userFactory
+     * @param RoleCollectionFactory $_roleCollectionFactory
+     */
+    public function __construct(
+        \Magento\Framework\Message\ManagerInterface $_messageManager,
```
remstone7: just some nit pick but could we alphabetize all these things? Ex. magento things then klaviyo things?

Setup/InstallData.php
```diff
@@ -0,0 +1,77 @@
...
+        $role->setName(self::KLAVIYO_ROLE_NAME)
+                ->setPid(0)
+                ->setRoleType(RoleGroup::ROLE_TYPE) 
+                ->setUserType(UserContextInterface::USER_TYPE_ADMIN);
+        $role->save();
+ 
+        $resource = [
+            'Magento_Backend::all'
+        ];
+        try{
```
remstone7:
```suggestion
        try {
```

Helper/Data.php
njparadis: good question, this change was part of my variable name cleanup efforts and I had left the original structure intact. you could make an argument that this is a bit more readable, but I have no problems with changing it

remstone7: i mean maybe just rename the original $storeId variable to $scopedStoreCode?  then just use that?

Helper/Data.php
```diff
@@ -49,17 +56,40 @@ protected function getScopeSetting($path){
...
+            $storeId = $this->_request->getParam('store');
+            $websiteId = $this->_request->getParam('website');
+        } else {
+            // In frontend area. Only concerned with store for frontend.
+            $storeId = $this->_storeId;
+        }
+
+        if (isset($storeId)) {
+            $scope = \Magento\Store\Model\ScopeInterface::SCOPE_STORE;
+            $scopeCode = $storeId;
```
remstone7: same ordeal here

Helper/Data.php
```diff
@@ -81,6 +111,36 @@ public function getPrivateApiKey()
...
     }
 
+    public function getKlaviyoUsername()
+    {
+        return $this->getScopeSetting(self::KLAVIYO_USERNAME);
+    }
+
+    public function unsetKlaviyoUsername() 
+    {
+        return $this->setScopeSetting(self::KLAVIYO_USERNAME, "klaviyo");
```
remstone7: can we use self::NAME

Helper/ListOptions.php
```diff
@@ -22,14 +22,14 @@ public function toOptionArray()
         // field in Klaviyo\Reclaim\Block\System\Config\Form\Field\Newsletter, so we pass
         // it over in the options array.
 
-        if (!$this->data_helper->getPrivateApiKey()) {
+        if (!$this->_dataHelper->getPrivateApiKey()) {
             return [[
                 'label' => 'To sync newsletter subscribers to Klaviyo, first save a <strong>Private Klaviyo API Key</strong> on the "General" tab.',
                 'value' => 0
```
smoucka: we're using 'label' and 'value' strings all over this sucker. could constant those bad boys

Observer/NewsletterSubscribeObserver.php
```diff
@@ -7,21 +7,21 @@
 
 class NewsletterSubscribeObserver implements ObserverInterface
 {
-    protected $data_helper;
+    protected $_dataHelper;
```
smoucka: this actually isn't a convention in php since you have the ability to set visibility.

smoucka: ahhh, I guess never mind. probably better to just be consistent since we're already using that convention https://github.com/klaviyo/magento2-klaviyo/pull/54/files#diff-735f70c1f948ebbd90915b1548d93432R12

PR #55, opened 2020-01-13
Fix JS error on checkout page when Klaviyo module disabled

ihor-sviziev:
When we disabled Klaviyo extension in the admin - we started getting following JS error on checkout page while editing email address field:
image

I investigated this issue and found that there is no check if Klaviyo is available. 

In addition to that - Klaviyo still saves email to quote, even when module is not enabled.

My PR checks if Klaviyo global variable (`window._learnq`) is not available - do not try to track or send requests about saving email to quote.

### DIFF
--- a/view/frontend/web/js/view/checkout/email.js
+++ b/view/frontend/web/js/view/checkout/email.js
@@ -34,6 +34,10 @@ define([
       var self = this;
       console.log('Klaviyo_Reclaim - Binding to #customer-email');
       jQuery('#maincontent').delegate('#customer-email', 'change', function (event) {
+        if (!window._learnq || !window._learnq.identify) {
+          return;
+        }
+        
         self._email = jQuery(this).val();
         if (!window._learnq.identify().email) {
           window._learnq.push(['identify', {
--- a/view/frontend/web/js/view/checkout/email.js
+++ b/view/frontend/web/js/view/checkout/email.js
@@ -29,12 +29,15 @@ define([
         return true;
       }
     },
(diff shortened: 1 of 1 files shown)


view/frontend/web/js/view/checkout/email.js
```diff
@@ -34,6 +34,10 @@ define([
       var self = this;
       console.log('Klaviyo_Reclaim - Binding to #customer-email');
       jQuery('#maincontent').delegate('#customer-email', 'change', function (event) {
+        if (!window._learnq || !window._learnq.identify) {
```
remstone7: looks good, just one comment

could we put this before the if/else in the `handleCheckout`? that way we don't get into the postUserEmail method?
Preferable something along the lines of:

```
handleCheckout: function(){
  if (!this.isKlaviyoActive()) {
    return;
  }
  ....
}

isKlaviyoActive: function(){
    if (!window._learnq || !window._learnq.identify) {
      return;
    }
    return true;
}
```

ihor-sviziev: Klaviyo js initialize asynchronously, so during initialization it might be still not available

remstone7: that's fine, still make the function isKlaviyoActive please and will ship with #54

ihor-sviziev: Added isKlaviyoActive function, but kept it inside "delegate" function, in order not to brake anything

Issue #19, opened 2019-01-31
Magento2.1.6 Klaviyo - Asymmetric transaction rollback.


mageprashant:
I have used your extension version 1.0.5 in Magento2.1.6.
**Go to Admin > Stores > Configuration > KLAVIYO > General > Settings**

- Enable Klaviyo Extension > Yes

- Public Klaviyo API Key > XXXXX

- Private Klaviyo API Key > XXXXX

after Save Config

go to frontend and I try to Newsletter Subscription and getting an error > https://www.screencast.com/t/T24Ncq3T

Thank you

alisovtsev:
In this case subcribtion creates without customer creation, but UserProfileNewsletterSubscribeObserver of Klaviyo\Reclaim request subscription customer by customerId, which generate exception of NoSuchEntity and then happens Asymmetric transaction error. So you may fix like this:
instead of 
`if ($subscriber->isStatusChanged()) {
          $customer = $this->customer_repository_interface->getById($subscriber->getCustomerId());

          if ($subscriber->isSubscribed()) {
            $this->data_helper->subscribeEmailToKlaviyoList(
                $customer->getEmail(),
                $customer->getFirstname(),
                $customer->getLastname()
            );
          } else {
            $this->data_helper->unsubscribeEmailFromKlaviyoList($customer->getEmail());
          }
        }`

replace to :
`if ($subscriber->isStatusChanged()) {
            if ($subscriber->getCustomerId()) {
              $object = $this->customer_repository_interface->getById($subscriber->getCustomerId());
            } else {
              $object = $subscriber;
            }

            if ($subscriber->isSubscribed()) {
                $this->data_helper->subscribeEmailToKlaviyoList(
                    $object->getEmail(),
                    $object->getFirstname(),
                    $object->getLastname()
                );
            } else {
                $this->data_helper->unsubscribeEmailFromKlaviyoList($object->getEmail());
            }
        }`

PR #56, opened 2020-01-16
Fix the issue with a newsletter subscription

iGerchak:
**Issue:**
When adding a new newsletter to Magento via custom module, 
Observer [NewsletterSubscribeObserver](https://github.com/klaviyo/magento2-klaviyo/blob/master/Observer/NewsletterSubscribeObserver.php) not called, but after saving newsletter to the Magento calls another Observer [UserProfileNewsletterSubscribeObserver](https://github.com/klaviyo/magento2-klaviyo/blob/master/Observer/UserProfileNewsletterSubscribeObserver.php) and try to get a customer by ID, Magento throws Exception because a customer with ID=0 doesn't exist.
As a result, an event about a new newsletter doesn't send to Klaviyo.

**Related issue:** #19 

**Steps to reproduce:**
1. Configure "Welcome email" in the Klaviyo
2. Create a new module for subscribe newsletter programmatically: https://magento.stackexchange.com/questions/150252/how-to-create-newsletter-subscriber-programmatically-magento2
3. Try to subscribe to a newsletter.

**Expected result:**
Email what was subscribed should get "Welcome email" 

**Actual result:**
"Welcome email" doesn't send


Observer/UserProfileNewsletterSubscribeObserver.php
```diff
@@ -12,32 +12,36 @@ class UserProfileNewsletterSubscribeObserver implements ObserverInterface
...
     }
 
     public function execute(\Magento\Framework\Event\Observer $observer)
     {
-        if (!$this->data_helper->getEnabled()) return;
-
+        if (!$this->data_helper->getEnabled()) {
+            return;
+        }
+        $customer   = null;
```
remstone7:
```suggestion
        $customer = null;
```

Observer/UserProfileNewsletterSubscribeObserver.php
```diff
@@ -12,32 +12,36 @@ class UserProfileNewsletterSubscribeObserver implements ObserverInterface
...
-            $this->data_helper->subscribeEmailToKlaviyoList(
-                $customer->getEmail(),
-                $customer->getFirstname(),
-                $customer->getLastname()
-            );
-          } else {
-            $this->data_helper->unsubscribeEmailFromKlaviyoList($customer->getEmail());
-          }
+            if ($subscriber->getCustomerId()) {
+                $customer = $this->customer_repository_interface->getById($subscriber->getCustomerId());
```
remstone7: looks like we don't do anything if customer is null could we return early?

iGerchak: If customer is null we get the email from a subscriber

Observer/UserProfileNewsletterSubscribeObserver.php
```diff
@@ -12,32 +12,36 @@ class UserProfileNewsletterSubscribeObserver implements ObserverInterface
...
+            }
+
+            if ($subscriber->isSubscribed()) {
+                $this->data_helper->subscribeEmailToKlaviyoList(
```
klaviyojad: could we implement below as:

```
if($customer)
   $subscriber = $customer;
```

and then the below would just be:

```
$this->data_helper->subscribeEmailToKlaviyoList(
                $subscriber->getEmail(),
                $subscriber->getFirstname(),
                $subscriber->getLastname()
            );
```

iGerchak: `subscriber` and `customer`, are different objects and it will be not right using it as the same

Observer/NewsletterSubscribeObserver.php
```diff
@@ -2,26 +2,49 @@
...
-        $this->data_helper->subscribeEmailToKlaviyoList($email);
+        $customer = null;
+        $subscriber = $observer->getDataObject();
+
+        if ($subscriber->isStatusChanged()) {
+            if ($subscriber->getCustomerId()) {
+                $customer = $this->customerRepository->getById($subscriber->getCustomerId());
+            }
+
+            if ($subscriber->isSubscribed()) {
```
ihor-sviziev: Now this class handle subscriptions for both guests & customers, so we don't need 2nd observer anymore

ihor-sviziev:
@remstone7 could you look at this PR?

PR #57, opened 2020-01-22
Fix reporting Unsubscribe from List event

ihor-sviziev:
Switch to Klaviyo Lists v2 API endpoint 
This endpoint triggers "Unsubscribed from %ListName%" event

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -163,7 +163,6 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     {
         $listId = $this->getNewsletter();
         $optInSetting = $this->getOptInSetting();
-        $apiKey = $this->getPrivateApiKey();
 
         $properties = [];
         $properties['email'] = $email;
@@ -173,7 +172,7 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
 
         $propertiesVal = ['profiles' => $properties];
 
-        $path = self::LIST_V2_API . $listId . $optInSetting . "?api_key=" . $apiKey;
+        $path = self::LIST_V2_API . $listId . $optInSetting;
 
         try {
             $response = $this->sendApiRequest($path, $propertiesVal, 'POST');
@@ -191,11 +190,9 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     public function unsubscribeEmailFromKlaviyoList($email)
     {
         $listId = $this->getNewsletter();
-        $apiKey = $this->getPrivateApiKey();
 
         $path = self::LIST_V2_API . $listId . self::API_SUBSCRIBE;
         $fields = [
-            'api_key' => (string)$apiKey,
             'emails' => [(string)$email],
         ];
 
@@ -249,6 +246,9 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
     private function sendApiRequest(string $path, array $params, string $method = 'POST')
     {
         $url = self::KLAVIYO_HOST . $path;
+
+        //Add API Key to params
+        $params['api_key'] = $this->getPrivateApiKey();
 
         $curl = curl_init();
         $encodedParams = json_encode($params);
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -178,24 +178,36 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
         return $response;
     }
 
-    public function unsubscribeEmailFromKlaviyoList($email) 
+    public function unsubscribeEmailFromKlaviyoList($email)
     {
         $list_id = $this->getNewsletter();
         $api_key = $this->getPrivateApiKey();
 
-        $fields = [
-            'api_key=' . $api_key,
-            'email=' . urlencode($email),
-        ];
+        $url = 'https://a.klaviyo.com/api/v2/list/' . $list_id . '/subscribe';
+        $fields = json_encode([
+            'api_key' => (string)$api_key,
+            'emails' => [(string)$email],
+        ]);
 
-        $ch = curl_init();
-        curl_setopt($ch, CURLOPT_URL, 'https://a.klaviyo.com/api/v1/list/' . $list_id . '/members/exclude');
-        curl_setopt($ch, CURLOPT_POST, count($fields));
-        curl_setopt($ch, CURLOPT_POSTFIELDS, join('&', $fields));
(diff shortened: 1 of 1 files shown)


Helper/Data.php
```diff
@@ -178,24 +178,36 @@ public function subscribeEmailToKlaviyoList($email, $first_name=null, $last_name
...
+        ]);
 
-        $ch = curl_init();
-        curl_setopt($ch, CURLOPT_URL, 'https://a.klaviyo.com/api/v1/list/' . $list_id . '/members/exclude');
-        curl_setopt($ch, CURLOPT_POST, count($fields));
-        curl_setopt($ch, CURLOPT_POSTFIELDS, join('&', $fields));
-        curl_setopt($ch, CURLOPT_RETURNTRANSFER, 1);
+        $curl = curl_init();
+        curl_setopt_array($curl, [
+            CURLOPT_URL => $url,
```
remstone7: this is okay and nothing wrong as it's just using v2... but i would much prefer a full refactor to create an isolated method that would handle making api requests, you would pass some arguments like `path, method, body` and then we wouldn't need to use curl all over the place like it currently is

ihor-sviziev:
@remstone7 I added fix.
I wanted to migrate to some magento class that doing what you need - but it doesn't support DELETE requsts O_o 
https://github.com/magento/magento2/pull/25742

Helper/Data.php
```diff
@@ -228,4 +237,44 @@ protected function make_request($path, $params) {
...
+            CURLOPT_CUSTOMREQUEST => $method,
+            CURLOPT_POSTFIELDS => $encodedParams,
+            CURLOPT_USERAGENT => self::USER_AGENT,
+            CURLOPT_HTTPHEADER => [
+                'Content-Type: application/json',
+                'Content-Length: ' . strlen($encodedParams)
+            ],
+        ]);
+
+        // Submit the POST request
```
remstone7: could we remove this comment or just make it:
```suggestion
        // Submit the request
```

ihor-sviziev: Sure

ihor-sviziev: Done

Helper/Data.php
ihor-sviziev: Fixed

Helper/Data.php
```diff
@@ -142,60 +151,60 @@ public function getKlaviyoLists($api_key=null){
...
+     * @return bool|string
+     */
+    public function unsubscribeEmailFromKlaviyoList($email)
     {
-        $list_id = $this->getNewsletter();
-        $api_key = $this->getPrivateApiKey();
+        $listId = $this->getNewsletter();
+        $apiKey = $this->getPrivateApiKey();
 
+        $path = 'api/v2/list/' . $listId . '/subscribe';
```
remstone7: we have a const for /subscribe, please use that

Helper/Data.php
```diff
@@ -142,60 +151,60 @@ public function getKlaviyoLists($api_key=null){
...
-        if ($last_name) $properties['$last_name'] = $last_name;
+        if ($firstName) $properties['$first_name'] = $firstName;
+        if ($lastName) $properties['$last_name'] = $lastName;
         if ($source) $properties['$source'] = $source;
 
-        $properties_val = count($properties) ? json_encode(array('profiles' => $properties)) : '{}';
+        $propertiesVal = ['profiles' => $properties];
 
-        $url = "https://a.klaviyo.com/api/v2/list/" . $list_id . $opt_in . "?api_key=" . $api_key;
+        $path = "api/v2/list/" . $listId . $optInSetting . "?api_key=" . $apiKey;
```
remstone7: one more just nit thing, could we make a const

`LIST_V2_API = 'api/v2/list/'`
then use `$path = self::LIST_V2_API . $listId.....`

Helper/Data.php
ihor-sviziev: OK

Helper/Data.php
ihor-sviziev: Fixed

Helper/Data.php
ihor-sviziev: Fixed

ihor-sviziev:
In general I would like to extract all logic that sends requests to API to some model(s), but it definitely would be separate PR. Now my need - fix the issue

Helper/Data.php
```diff
@@ -142,60 +152,60 @@ public function getKlaviyoLists($api_key=null){
...
-        $list_id = $this->getNewsletter();
-        $api_key = $this->getPrivateApiKey();
+        $listId = $this->getNewsletter();
+        $apiKey = $this->getPrivateApiKey();
 
+        $path = self::LIST_V2_API . $listId . self::API_SUBSCRIBE;
         $fields = [
-            'api_key=' . $api_key,
-            'email=' . urlencode($email),
+            'api_key' => (string)$apiKey,
```
remstone7: sorry one more thing (but looks good 👍), i know our api accepts two different types of auth either utm param of api_key or api_key in the json, could we move this into the sendApiRequest and add it as a param?

ihor-sviziev: Why we need it? I did it as showed in the docs example:
https://www.klaviyo.com/docs/api/v2/lists#delete-subscribe
image

Is there any issues or benefits of doing that?

ihor-sviziev: Or you meant that we can add api key param inside `sendApiRequest`,
like 
```php
private function sendApiRequest(string $path, array $params, string $method = 'POST')
    {
        $params['api_key'] = $this->getPrivateApiKey(); // <!-- see to this line
        $url = self::KLAVIYO_HOST . $path;

        $curl = curl_init();
        ...

remstone7: yes please to this one 👍

remstone7:
just 1 conflict should be a quick fix -- also, we're just bringing this down to test today/tomorrow and will merge with out 1.1.7 release

Helper/Data.php
ihor-sviziev: Done

ihor-sviziev:
@remstone7 I fixed conflict

ihor-sviziev:
@remstone7 I see that you already released version 1.1.7 and 1.1.8, did you missed to merge this PR?

remstone7:
Missed this but merging now, had an urgent fix we had to get out.  Itll go out in 1.1.9 which should be in the next couple weeks

PR #59, opened 2020-02-06
update namespace used in config.xml

njparadis:
* Let Magento know to decrypt the password before creating a new user


ogsmith:
lgtm

PR #63, opened 2020-02-07
Make our own logger; add API endpoints to query/clean log file

njparadis:
Setup a Logger class to log out to out own klaviyo.log file under /var/log
Replaced instances where we use the built-in Magento logger with our own

TODO:
* ~~Appropriate permissions level for the API endpoint (I think Magento_Backend::Admin but need to research)~~
* Expand the stuff we log out
    * Newsletter subscriptions?
    * Newsletter unsubscribes?
    * Other exceptions?
* ~~Figure out how to retrieve the log file contents in the getLog() function~~


Logger/Logger.php
```diff
@@ -0,0 +1,6 @@
+<?php
+namespace Klaviyo\Reclaim\Logger;
+
+class Logger extends \Monolog\Logger
```
remstone7: should this be called KlaviyoLogger?

Setup/InstallData.php
```diff
@@ -26,11 +26,11 @@ class InstallData implements InstallDataInterface
...
     /**
-     * Logger
-     *
-     * @var logger
+     * Logging instance
+     * 
+     * @var \Klaviyo\Reclaim\Logger\Logger
      */
-    protected $logger;
+    protected $_logger;
```
remstone7: could you call it $_klaviyoLogger ? (or whatever case we use)

etc/di.xml
```diff
@@ -2,4 +2,17 @@
 <config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:ObjectManager/etc/config.xsd">
     <preference for="Klaviyo\Reclaim\Api\ReclaimInterface"
                 type="Klaviyo\Reclaim\Model\Reclaim" />
+    <type name="Klaviyo\Reclaim\Logger\Handler">
```
jordanallain: any way to pull these string literals out into a constant? at least for the ones that are repeated?

remstone7: eh, it's okay to leave it imo, it's a magento standard, we shouldn't really deviate from it

Logger/Logger.php
njparadis: It can be, it is under the \Klaviyo\Reclaim\Logger namespace already so it feels a bit redundant to me.

njparadis: Since it would look like this wherever we use it:
\Klaviyo\Reclaim\Logger\KlaviyoLogger

Model/Reclaim.php
```diff
@@ -60,7 +88,99 @@ public function reclaim()
...
+     */
+    public function cleanLog($date)
+    {
+        //get log file path and do the old switcheroo in preparation for cleaning
+        $path = $this->_dir->getPath('log') . '/klaviyo.log';
+        $old = $path . ".old";
+        rename($path, $old);
+
+        //open file streams
+        $input = fopen($old, "rb");
```
remstone7: only bad thing about files in php is it takes up a heap of memory.  is there a more efficient way to do this? we should thing about how actively we need to clean this / how to prevent us from reading a file that could be too big

njparadis: is this still a concern if we are using a filestream? my understanding is that this would only be loading one line of the file into memory at a time.

could be an issue in getLog fwiw, since that method does just load the entire contents of the file into memory. unsure how else we would be able to return a response without doing that though

Setup/Uninstall.php
```diff
@@ -0,0 +1,34 @@
...
+use Magento\Framework\Setup\ModuleContextInterface;
+use Magento\Framework\Filesystem\DirectoryList;
+
+class Uninstall implements UninstallInterface
+{
+    /**
+     * DirectoryList instance
+     * @var \Magento\Framework\Filesystem\DirectoryList $_dir
+     */
+    protected $_dir;
```
jordanallain: Is this used? Maybe my own ignorance of PHP but I see `_dir` and `$dir` used but not `$_dir`

njparadis: the $_dir on this line is the same as $this->_dir referenced below; $dir is the name of the variable passed into the constructor

Model/Reclaim.php
```diff
@@ -47,10 +75,114 @@ public function __construct(
...
 
+    /**
+     * Returns the Klaviyo log file
+     * 
+     * @api
+     * @return string
+     */
+    public function getLog()
+    {
+        $log = file($this->_dir->getPath('log') . '/klaviyo.log');
```
remstone7: are we able to just use the `class Handler extends \Magento\Framework\Logger\Handler\Base
` to get this? seems like it's the same thing?

remstone7: nvm

Setup/InstallData.php
```diff
@@ -26,35 +33,46 @@ class InstallData implements InstallDataInterface
...
-        $this->logger = $logger;
+        $this->_klaviyoLogger = $klaviyoLogger;
     }
  
     public function install(
         ModuleDataSetupInterface $setup, 
         ModuleContextInterface $context
     ) {
+        //Klaviyo log file creation
+        $path = $this->_dir->getPath('log') . '/klaviyo.log';
```
remstone7: i feel like this is kinda reused all of the place, the getting of the path and such, is this able to be isolated?

Setup/UpgradeData.php
```diff
@@ -0,0 +1,36 @@
...
+        $this->_dir = $dir;
+    }
+
+    public function upgrade(
+        ModuleDataSetupInterface $setup,
+        ModuleContextInterface $context
+    )
+    {
+        //Klaviyo log file creation
+        $path = $this->_dir->getPath('log') . '/klaviyo.log';
```
remstone7: it's everywhere!

etc/di.xml
```diff
@@ -2,4 +2,17 @@
...
+        <arguments>
+            <argument name="filesystem" xsi:type="object">Magento\Framework\Filesystem\Driver\File</argument>
+        </arguments>
+    </type>
+    <type name="Klaviyo\Reclaim\Logger\Logger">
```
remstone7: should logging be on by default? I would think it's not a bad idea but maybe customers want to turn it off?

Setup/InstallData.php
```diff
@@ -26,35 +33,46 @@ class InstallData implements InstallDataInterface
...
      */
-    protected $logger;
+    protected $_klaviyoLogger;
  
     const KLAVIYO_ROLE_NAME = 'Klaviyo';
 
     /**
      * Init
      *
+     * @param DirectoryList $dir
```
remstone7: nice!

Model/Reclaim.php
```diff
@@ -47,10 +75,114 @@ public function __construct(
...
+     * @return mixed[]
+     */
+    public function appendLog($message)
+    {
+        //log the provided message
+        $this->klaviyoLogger->info($message);
+
+        //return success message
+        return array(
+            "message" => 'Logged message: \'' . $message . '\''
```
remstone7: could we standardize single or double quotes

Model/Reclaim.php
```diff
@@ -47,10 +75,114 @@ public function __construct(
...
+            $response = array(
+                'message' => 'Unable to parse timestamp: ' . $date
+            );
+            $this->klaviyoLogger->info("cleanLog failed: unable to parse timestamp from: " . $date);
+            return $response;
+        }
+        
+        //get log file path and do the old switcheroo in preparation for cleaning
+        $path = $this->_dir->getPath('log') . '/klaviyo.log';
+        $old = $path . ".old";
```
remstone7: so we rename the filt to klaviyo.log.old?

Model/Reclaim.php
```diff
@@ -47,10 +75,114 @@ public function __construct(
...
+        rename($path, $old);
+
+        //open file streams
+        $input = fopen($old, "rb");
```
remstone7: so if i get this right the logic is:

1. when we go to clean the log, we make a copy
2. we create a new log file, then for everything in the old file that the timestamp is > the $date we will write to the new file
3. purge the old file?

njparadis: correct, should we be doing something different?

Setup/UpgradeData.php
```diff
@@ -4,33 +4,83 @@
...
             fopen($path, 'w');
         }
         chmod($path, 0644);
+
+        /**
+         * in release 1.1.7 we started using the encrypted backend model for the private api key
+         * this check ensures that when upgrading to this version the key is stored properly
+         * otherwise we'd be trying to decrypt an unencrypted value elsewhere in the extension code (yikes)
+         */
+        if (version_compare($context->getVersion(), '1.1.7', '<')) {
```
klaviyojad, line 76: can we make that a const?

Helper/ScopeSetting.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    /**
+     * helper function to allow this class to be used in Setup files
+     */
+    protected function checkAreaCode()
+    {
+        //when this class is accessed from cli commands, there is no area code set
+        //(since there is no actual session running persay)
+        //this try-catch block is needed to allow this helper to be used in setup files
```
smoucka, line 52: Nit: can we just change this to a multi-line comment?

Setup/UpgradeData.php
njparadis: the version # ? or something else

njparadis: due to the way that PHP's memory management works (it's Not Great) I've been trying to avoid const-ing strings unless I use it more than once

Helper/ScopeSetting.php
```diff
@@ -0,0 +1,195 @@
...
+    {
+        return $this->getScopeSetting(self::PRIVATE_API_KEY);
+    }
+
+    public function setPrivateApiKey($value)
+    {
+        return $this->setScopeSetting(self::PRIVATE_API_KEY, $value);
+    }
+
+    public function getLoggerEnabled()
```
smoucka, line 142: maybe rename this function to isLoggerEnabled if this is going to return a boolean.

Helper/ScopeSetting.php
```diff
@@ -0,0 +1,195 @@
...
+        };
+    }
+
+    public function getVersion()
+    {
+        return $this->_moduleList
+            ->getOne(self::MODULE_NAME)['setup_version'];
+    }
+
+    public function getEnabled()
```
smoucka, line 122: same as below, could rename to isEnabled

Helper/ScopeSetting.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    public function getKlaviyoPassword()
+    {
+        return $this->getScopeSetting(self::KLAVIYO_PASSWORD);
+    }
+
+    public function unsetKlaviyoPassword()
+    {
+        return $this->setScopeSetting(self::KLAVIYO_PASSWORD, "");
```
smoucka, line 164: nit: single quotes

Helper/ScopeSetting.php
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    public function getKlaviyoEmail()
+    {
+        return $this->getScopeSetting(self::KLAVIYO_EMAIL);
+    }
+
+    public function unsetKlaviyoEmail()
+    {
+        return $this->setScopeSetting(self::KLAVIYO_EMAIL, "");
```
smoucka, line 174: single quotes

Helper/Logger.php
```diff
@@ -0,0 +1,75 @@
...
+    public function log($message)
+    {
+        if ($this->loggerEnabled) {
+            $this->_klaviyoLogger->info($message);
+        }
+    }
+
+    /**
+     * do I need this? who tf knows
+     * probably
```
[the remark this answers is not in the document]
njparadis: image

njparadis:
image

njparadis:
image

etc/webapi.xml
```diff
@@ -48,4 +48,22 @@
             <resource ref="anonymous"/>
         </resources>
     </route>
+    <route url="/V1/klaviyo/reclaim/log" method="GET">
```
ihor-sviziev, line 51: Hi @njparadis,
Could you explain why it could be needed - managing logs via REST API?
It looks to me as security issue because in logs might be cleaned via rest api.

njparadis: These methods only add or remove messages from our own custom log file added in this PR. By default the logger is disabled.

The idea behind these endpoints is to help debug issues with the extension without needing to wait for the customer to send over the log files. The clean method will be called as part of the periodic sync our app performs to ensure that the log file does not grow to be too large.

PR #74, opened 2020-07-07
Allow URL parameters to persist through cart redirect

caroline-d-lu:
If the URL has parameters other than the quote ID, allow those parameters to persist through the cart redirect. If there are no other parameters in the URL, redirect normally.

### DIFF
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -43,7 +43,7 @@ class Cart extends \Magento\Framework\App\Action\Action
         }
 
         $redirect = $this->resultRedirectFactory->create();
-        $redirect->setPath('checkout/cart',['_query'=>$params]);
+        $redirect->setPath('checkout/cart', ['_query'=>$params]);
         return $redirect;
     }
 }
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -43,11 +43,7 @@ class Cart extends \Magento\Framework\App\Action\Action
         }
 
         $redirect = $this->resultRedirectFactory->create();
-        if (sizeof($params) > 0) {
-            $redirect->setPath('checkout/cart',['_query'=>$params]);
-        } else {
-            $redirect->setPath('checkout/cart');
-        }
+        $redirect->setPath('checkout/cart',['_query'=>$params]);
         return $redirect;
     }
 }
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -32,6 +32,8 @@ class Cart extends \Magento\Framework\App\Action\Action
     public function execute()
     {
         $quoteId = $this->request->getParam('quote_id');
+        $params = $this->request->getParams();
+        unset($params['quote_id']);
 
         try {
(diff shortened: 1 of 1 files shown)


Controller/Checkout/Cart.php
```diff
@@ -32,6 +32,8 @@ public function __construct(
     public function execute()
     {
         $quoteId = $this->request->getParam('quote_id');
```
remstone7, line 34: we shouldn't need to call the request getParam and getParams, this can be cleaned up.

Controller/Checkout/Cart.php
```diff
@@ -41,7 +43,11 @@ public function execute()
         }
 
         $redirect = $this->resultRedirectFactory->create();
-        $redirect->setPath('checkout/cart');
+        if (sizeof($params) > 0) {
+            $redirect->setPath('checkout/cart',['_query'=>$params]);
```
remstone7, line 47:
```suggestion
            $redirect->setPath('checkout/cart', ['_query'=>$params]);
```

remstone7: what happens if params is null/has no params and is passed into this as the second arg?

caroline-d-lu: good point -- should I change to if ($params != NULL && sizeof($params) > 0)  ?

Controller/Checkout/Cart.php
caroline-d-lu: do you mean call $this->request->getParams() once and then grab quoteId and then unset quoteId?

remstone7: yea that would be great

Controller/Checkout/Cart.php
remstone7: no i'm asking what happens if params is null and get's passed into the setPath function, is the if/else necessary?

caroline-d-lu: Ohhh I get what you're saying now

caroline-d-lu: @kamil-klasicki what gets returned by "['_query'=>$params]" if $params is null? Also why are we doing "['_query'=>$params]" instead of just passing $params?

klaviyojad:
Looks good but aren't we going to update the version of the extension as well as the readme?

Controller/Checkout/Cart.php
```diff
@@ -31,7 +31,9 @@ public function __construct(
      */
     public function execute()
     {
-        $quoteId = $this->request->getParam('quote_id');
+        $params = $this->request->getParams();
+        $quoteId = $params['quote_id'];
```
ihor-sviziev, line 35: if `quote_id` isn't exists - we'll have there notice. 
```suggestion
        $quoteId = $params['quote_id'] ?? 0;
```

remstone7:
Quote I'd will always be there, this is a custom link that we build and is in our default abandoned cart emails.  We can add the safety net in a future release or feel free to make a pr!

Controller/Checkout/Cart.php
remstone7: https://github.com/klaviyo/magento2-klaviyo/pull/76

PR #76, opened 2020-07-14
defensive cart rebuild

remstone7:


Controller/Checkout/Cart.php
```diff
@@ -32,7 +32,7 @@ public function __construct(
     public function execute()
     {
         $params = $this->request->getParams();
-        $quoteId = $params['quote_id'];
+        $quoteId = $params['quote_id'] ?? 0;
```
smoucka, line 35: this requires php >=7. i don't know if we have people using pre-7 or not

smoucka: If we do still want to support 5x we should use isset with a ternary - https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op

remstone7: yea, just checked looks like 2.0* works with 5.6+ but 2.1+ needs php 7*.. i'll update just incase we have some on 5.*

ihor-sviziev: I would add following link https://www.php.net/supported-versions.php - all version before php 7.2 already NOT supported officially and not getting security updates, which is critical for ecommerce websites to be secure. I support of not supported versions could be just removed.

PR #79, opened 2020-10-09
Remove unused variable, CSP report-only mode

njparadis:
- Removes unused variable and DI from Reclaim.php
- Update CSP to be in report-only mode
- Bump version to 1.2.3

### DIFF
--- a/etc/config.xml
+++ b/etc/config.xml
@@ -12,5 +12,15 @@
                 <password backend_model="Magento\Config\Model\Config\Backend\Encrypted" />
             </klaviyo_user>
         </klaviyo_reclaim_user>
+                <csp>	
+            <mode>	
+                <storefront>	
+                    <report_only>1</report_only>	
+                </storefront>	
+                <admin>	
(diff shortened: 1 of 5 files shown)


etc/config.xml
```diff
@@ -12,5 +12,15 @@
                 <password backend_model="Magento\Config\Model\Config\Backend\Encrypted" />
             </klaviyo_user>
         </klaviyo_reclaim_user>
+                <csp>
```
remstone7, line 15: is 1 the default? i take it as last pr we did 0?

njparadis: Yeah I had a couple of customers reach out about issues they ran into with the CSP settings in 1.2.2; I had initially just removed it based on conversion with first customer, but after a more recent ticket/convo and additional research, I think this approach is better.

The tldr is enabling report-only mode shouldn't effect other modules, but disabling it can cause issues - this explicitly enables report-only mode, rather than relying on default behavior.

remstone7 (approved):

PR #81, opened 2020-10-14
Removed CSP setting

jrodriguez-ifuelinteractive:
Setting CSP settings in an extension not maintained by a merchant can cause un-intended issues. Leaving the CSP settings up to the merchant is a better approach.


njparadis (approved):

njparadis:
Merging alongside #90 to be packaged in next release

PR #83, opened 2020-11-20
Remove jQuery UI to improve performance

iGerchak:
jQuery UI is never used in the email.js, but including it on the **define**, and it starts loading the all UI library (jQuery UI with all Widgets).
As a result, email.js will be run just after all libs with included in **define**.
Also, it creates a performance issue for the checkout page.


remstone7 (approved):
👍

ihor-sviziev (approved):

PR #88, opened 2020-11-30
Product Delete Webhook & API

remstone7:
This does the first pass for starting to add webhooks.

The first one will be the ProductDeleteBefore as well as adding a webhook secret for validation

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -162,9 +162,9 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      * @param string $path
      * @param array $params
      * @param string $method
-     * @return bool|string
      * @throws \Exception
      */
+
     private function sendApiRequest(string $path, array $params, string $method = null)
     {
         $url = self::KLAVIYO_HOST . $path;
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -164,7 +164,6 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      * @param string $method
      * @throws \Exception
      */
-
     private function sendApiRequest(string $path, array $params, string $method = null)
     {
         $url = self::KLAVIYO_HOST . $path;
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -162,6 +162,7 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      * @param string $path
      * @param array $params
      * @param string $method
+     * @return mixed[]
      * @throws \Exception
      */
     private function sendApiRequest(string $path, array $params, string $method = null)
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -21,7 +21,7 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     const KLAVIYO_EMAIL = 'klaviyo_reclaim_user/klaviyo_user/email';
 
     const WEBHOOK_SECRET = 'klaviyo_reclaim_webhook/klaviyo_webhooks/webhook_secret';
-    const PRODUCT_DELETE_AFTER = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_delete_after_webhook';
+    const PRODUCT_DELETE_BEFORE = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_delete_before_webhook';
 
     protected $_scopeConfig;
     protected $_request;
@@ -207,11 +207,12 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
         }
     }
 
-    /** This maps a klaviyo account to all the store ids its scoped too.
+    /**
+     * This maps a klaviyo account to all the store ids its scoped too.
      * @param $storeIds
      * @return array
      */
-    public function getStoredIdKlaviyoAccountSetMap($storeIds)
+    public function getStoreIdKlaviyoAccountSetMap($storeIds)
     {
 
         $storeMap = array();
@@ -227,9 +228,9 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
         return $storeMap;
     }
 
-    public function getProductDeleteAfterSetting($storeId = null)
+    public function getProductDeleteBeforeSetting($storeId = null)
     {
-        return $this->getScopeSetting(self::PRODUCT_DELETE_AFTER, $storeId);
+        return $this->getScopeSetting(self::PRODUCT_DELETE_BEFORE, $storeId);
     }
 
 }
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -21,7 +21,7 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     const KLAVIYO_EMAIL = 'klaviyo_reclaim_user/klaviyo_user/email';
 
     const WEBHOOK_SECRET = 'klaviyo_reclaim_webhook/klaviyo_webhooks/webhook_secret';
-    const PRODUCT_DELETE_AFTER = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_update_webhook';
+    const PRODUCT_DELETE_AFTER = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_delete_after_webhook';
 
     protected $_scopeConfig;
     protected $_request;
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -224,7 +224,6 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
                 array_push( $storeMap[$klaviyoAccount], $storeId);
             }
         }
-
         return $storeMap;
     }
 
@@ -232,6 +231,4 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     {
         return $this->getScopeSetting(self::PRODUCT_DELETE_BEFORE, $storeId);
     }
-
 }
-
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -73,7 +73,7 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     {
         $this->checkAreaCode();
 
-        if ($storeId) {
+        if (isset($storeId)) {
             $scopedStoreCode = $storeId;
         } elseif ($this->_state->getAreaCode() == \Magento\Framework\App\Area::AREA_ADMINHTML) {
             $scopedStoreCode = $this->_request->getParam('store');
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -228,7 +228,7 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
         return $storeMap;
     }
 
-    public function getProductDeleteBeforeSetting($storeId = null)
+    public function getProductDeleteBeforeSetting($storeId)
     {
         return $this->getScopeSetting(self::PRODUCT_DELETE_BEFORE, $storeId);
     }
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -20,6 +20,9 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     const KLAVIYO_PASSWORD = 'klaviyo_reclaim_user/klaviyo_user/password';
     const KLAVIYO_EMAIL = 'klaviyo_reclaim_user/klaviyo_user/email';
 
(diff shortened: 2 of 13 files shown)


Helper/Webhook.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * @param array data
+     * @return string
+     * @throws Exception
+     */
+    private function createWebhookSecurity(array $data)
+    {
+        $webhookSecret = $this->_klaviyoScopeSetting->getWebhookSecret();
+        return hash_hmac('sha256', json_encode($data), $webhookSecret);
```
klaviyojad, line 74: is this a standard approach?  do we do this everywhere else we do webhooks.  It would be great to document why you do it like that

etc/adminhtml/events.xml
```diff
@@ -9,4 +9,7 @@
     name="admin_system_config_changed_section_klaviyo_reclaim_user">
         <observer name="custom_admin_system_config_changed_section_klaviyo_reclaim_user" instance="Klaviyo\Reclaim\Observer\KlaviyoUserObserver"/>
     </event>
+    <event name="catalog_product_delete_after">
```
klaviyojad, line 12: It would be nice to document that the above is a standard thing that gets called in magento2 for when a product is deleted https://magento.stackexchange.com/questions/300931/magento-2-delete-event-get-product-skus-using-observer

remstone7: this is a standard observer in magento, i think this is pretty self explanatory ?

Helper/Webhook.php
remstone7: yes it is standard, not going to document in a public repo as to why

composer.json
```diff
@@ -13,5 +13,8 @@
     },
     "config": {
         "optimize-autoloader": true
+    },
+    "require": {
+        "ext-curl": "*"
```
remstone7, line 18: i think this is already required in magento, but i feel like its not a bad thing to have since our extension depends on it

etc/adminhtml/system.xml
```diff
@@ -26,7 +26,7 @@
                     <label>Custom Media URL</label>
                     <validate>validate-url</validate>
                 </field>
-                <field id="logger" translate="label" type="select" sortOrder="50" showInDefault="1" showInWebsite="1" showInStore="1">
+                <field id="logger" translate="label" type="select" sortOrder="60" showInDefault="1" showInWebsite="1" showInStore="1">
```
remstone7, line 29: accident, will fix back to 50

etc/adminhtml/system.xml
```diff
@@ -71,5 +71,21 @@
...
+        <section id="klaviyo_reclaim_webhook" translate="label" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
+            <label>Webhooks</label>
+            <tab>klaviyo</tab>
+            <resource>Klaviyo_Reclaim::klaviyo_reclaim</resource>
+            <group id="klaviyo_webhooks" translate="label" type="text" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
+                <label>Setup Klaviyo Webhooks</label>
+                <field id="webhook_secret" translate="label" type="text" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
+                    <label>Webhook Secret</label>
+                </field>
+                <field id="using_product_update_webhook" translate="label" type="select" sortOrder="60" showInDefault="1" showInWebsite="1" showInStore="1">
```
remstone7, line 83: will make this 20 sortOrder

Helper/Webhook.php
```diff
@@ -0,0 +1,78 @@
...
+                'Content-Length: '. strlen(json_encode($data))
+            ),
+        ]);
+
+        // Submit the request
+        $response = curl_exec($curl);
+        $err = curl_errno($curl);
+
+        if ($err) {
+            throw new Exception(curl_error($curl));
```
remstone7, line 57: idk if we should do this or not, maybe just log it?

Helper/Webhook.php
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Klaviyo\Reclaim\Helper;
+
+use Exception;
+use \Klaviyo\Reclaim\Helper\ScopeSetting;
+
+class Webhook extends \Magento\Framework\App\Helper\AbstractHelper
+{
+    const USER_AGENT = 'Klaviyo/Webhook/1.0';
+    const WEBHOOK_URL = 'http://www.klaviyo.com/api/webhook/integration/magento_two';
```
remstone7, line 11: should be https

Helper/Webhook.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace Klaviyo\Reclaim\Helper;
+
+use Exception;
+use \Klaviyo\Reclaim\Helper\ScopeSetting;
+
+class Webhook extends \Magento\Framework\App\Helper\AbstractHelper
```
remstone7, line 8: should this be named Webhook or should it be more descriptive?

Observer/ProductDeleteAfter.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    protected $_objectManager;
+
+    /**
+     * @param \Klaviyo\Reclaim\Helper\Webhook $webhookHelper
+     * @param \Magento\Framework\ObjectManagerInterface $objectManager
+     * @param ScopeSetting $klaviyoScopeSetting
+     */
+    public function __construct(
+        \Klaviyo\Reclaim\Helper\Webhook $webhookHelper,
```
remstone7, line 33: ph man, i should clean this up, do we have a preference on injections?

Helper/ScopeSetting.php
```diff
@@ -20,6 +20,9 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     const KLAVIYO_PASSWORD = 'klaviyo_reclaim_user/klaviyo_user/password';
     const KLAVIYO_EMAIL = 'klaviyo_reclaim_user/klaviyo_user/email';
 
+    const WEBHOOK_SECRET = 'klaviyo_reclaim_webhook/klaviyo_webhooks/webhook_secret';
+    const PRODUCT_DELETE_AFTER = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_update_webhook';
```
klaviyojad, line 24: the const is named DELETE and yet the string has the name 'using_product_update_webhook' Its my understanding that this webhook will only fire after a product is deleted, not updated

Helper/ScopeSetting.php
```diff
@@ -121,6 +124,11 @@ public function getVersion()
             ->getOne(self::MODULE_NAME)['setup_version'];
     }
 
+    public function getWebhookSecret()
```
klaviyojad, line 127: how come there is no setWebhookSecret ?

Helper/Webhook.php
```diff
@@ -0,0 +1,78 @@
...
+{
+    const USER_AGENT = 'Klaviyo/Webhook/1.0';
```
klaviyojad, line 10: is this going to be the standard for all extensions?  Do you think we should also contain the name of the integration within it?  Like `Klaviyo/MagentoTwo/Webhook/1.0` ?

Helper/ScopeSetting.php
remstone7: good call, will rename 👍

Helper/ScopeSetting.php
remstone7: we don't ever need to set this value, its done through the system.xml file, we just need to retrieve it.

klaviyojad: Could you explain more? When I look at the below its not clear to me that I see any value

```
                <field id="webhook_secret" translate="label" type="password" sortOrder="10" showInDefault="1" showInWebsite="1" showInStore="1">
                    <label>Webhook Secret</label>
                    <backend_model>Magento\Config\Model\Config\Backend\Encrypted</backend_model>
                </field>
```

remstone7: it's done natively through magento, we don't have to do a set function for this.  we just need to be able to retrieve it.

Helper/Webhook.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param array data
+     * @return string
+     * @throws Exception
+     */
+    private function createWebhookSecurity(array $data)
+    {
+        $webhookSecret = $this->_klaviyoScopeSetting->getWebhookSecret();
```
remstone7, line 74: the only thing here is a may need to add in a store ID for the webhook secret because things can be scoped differently..

Helper/Webhook.php
```diff
@@ -0,0 +1,79 @@
...
+    const WEBHOOK_URL = 'https://rem-klaviyo.ngrok.io/test';
+
+    /**
+     * Klaviyo scope setting helper
+     * @var \Klaviyo\Reclaim\Helper\ScopeSetting $klaviyoScopeSetting
+     */
+    protected $_klaviyoScopeSetting;
+
+    public function __construct(
+        ScopeSetting $klaviyoScopeSetting
```
jordanallain, line 21: is there a reason for this being on its own line?

remstone7: yea, just how its styled

Helper/Webhook.php
jordanallain: like `KlaviyoWebhook` or something even more descriptive?

smoucka (commented):
Just some cleanup but otherwise looks pretty good.

One larger thought, and this could very well be outside the scope of this pr, but any merit in a base class for other observers to inherit from? I assume other product related webhooks are going to be similar in terms of store scoping.

Helper/Data.php
```diff
@@ -162,9 +162,9 @@ protected function make_request($path, $params) {
      * @param string $path
      * @param array $params
      * @param string $method
-     * @return bool|string
      * @throws \Exception
      */
+
     private function sendApiRequest(string $path, array $params, string $method = null)
```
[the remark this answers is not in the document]
remstone7: this is just my ide removing a buch of whitespace

remstone7:
> One larger thought, and this could very well be outside the scope of this pr, but any merit in a base class for other observers to inherit from? I assume other product related webhooks are going to be similar in terms of store scoping.

yes 100% this will happen, wasn't going to do it in this PR for time sensitivity, but we will have base classes for orders/products

Helper/Data.php
```diff
@@ -162,7 +162,6 @@ protected function make_request($path, $params) {
      * @param string $path
      * @param array $params
      * @param string $method
-     * @return bool|string
```
smoucka, line 165: nit: this method still returns something

Helper/ScopeSetting.php
```diff
@@ -194,4 +206,32 @@ public function getOptInSetting()
...
+        return $storeMap;
+    }
+
+    public function getProductDeleteBeforeSetting($storeId)
+    {
+        return $this->getScopeSetting(self::PRODUCT_DELETE_BEFORE, $storeId);
+    }
+
+}
+
```
smoucka, line 237: nit: i think we have two extra lines here? i believe gh doesn't show the newline at the end of file but i could be wrong

smoucka (approved):
lgtm. two nit, non-blocking formatting things

jordanallain (approved):

Helper/ScopeSetting.php
```diff
@@ -194,4 +206,29 @@ public function getOptInSetting()
             return self::API_MEMBERS;
         }
     }
-}
\ No newline at end of file
+
+    /**
+     * This maps a klaviyo account to all the store ids its scoped too.
```
jordanallain, line 211: small grammatical nit:

```diff
- This maps a klaviyo account to all the store ids its scoped too.
+ This maps a klaviyo account to all the store ids it's scoped to.
```

Model/Reclaim.php
```diff
@@ -247,7 +252,7 @@ public function productVariantInventory($product_id, $store_id=0)
         }
         // if store_id is specificed, use it
```
jordanallain, line 253: specified*

Helper/Webhook.php
jordanallain: did you decide against this?

remstone7: yea, just let it be global for now

jordanallain (approved):

Observer/ProductDeleteBefore.php
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @param Observer $observer
+     * @return void
+     * @throws Exception
+     */
+    public function execute(Observer $observer)
+    {
+        $product = $observer->getEvent()->getProduct();
+        $websiteIds = $product->getWebsiteIds();
+        $storeIdKlaviyoMap = $this->_klaviyoScopeSetting->getStoreIdKlaviyoAccountSetMap($websiteIds);
```
ihor-sviziev, line 58: Hi @remstone7,
There is a mistake. You're sending website IDs, while the method receives Store IDs.
Here are more details between them:
https://docs.magento.com/user-guide/configuration/scope.html

PR #90, opened 2020-12-02
Changelog to include changes in #81

njparadis:


njparadis:
Merging alongside #81

PR #93, opened 2020-12-03
Create OAuth Configuration

klaviyojad:
Updates to the Extension to expose a new Config Section to Setup the Integrations in Magento2 with the Identify and Redirect Oauth URLs


Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,96 @@
...
+
+    /*
+     * @param Observer $observer
+     * @return void
+     * @throws Exception
+     */
+    public function execute(\Magento\Framework\Event\Observer $observer)
+    {
+        if (empty($this->_klaviyoScopeSetting->getPublicApiKey())) {
+         throw new \Magento\Framework\Exception\StateException(__('To set up Oauth for Klaviyo, first save a <strong>Public Klaviyo API Key</strong> on the "General" tab.'));
```
remstone7, line 62: this isn't the correct language, you don't save a public key, i'd say they should get the public api key -- is it worth adding in the hc link on how to find it:

https://help.klaviyo.com/hc/en-us/articles/115005062267-Manage-Your-Account-s-API-Keys#your-public-api-key-site-id2

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,96 @@
...
+            $integration->setConsumerId($consumerId);
+            $integration->save();
+
+            // Code to grant permission
+            $this->_authorizationService->grantAllPermissions($integrationId);
+
+            //reset the details in the store config
+            $this->_klaviyoScopeSetting->unsetKlaviyoOauthName();
+        }catch(Exception $e){
+            throw new \Magento\Framework\Exception\StateException(__('Error creating Oauth Integration: '.$e->getMessage()));
```
remstone7, line 92:
```suggestion
            throw new \Magento\Framework\Exception\StateException(__('Error creating OAuth Integration: '.$e->getMessage()));
```

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,96 @@
...
+            $consumer = $this->_oauthService->createConsumer(['name' => $consumerName]);
+            $consumerId = $consumer->getId();
+            $integration->setConsumerId($consumerId);
+            $integration->save();
+
+            // Code to grant permission
+            $this->_authorizationService->grantAllPermissions($integrationId);
+
+            //reset the details in the store config
+            $this->_klaviyoScopeSetting->unsetKlaviyoOauthName();
```
remstone7, line 90: is there a reason we unset this? or does it just unset in the ui form field?

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public function __construct(
+        ScopeSetting $klaviyoScopeSetting,
+        MessageManager $messageManager,
+        IntegrationFactory $integrationFactory,
+        AuthorizationService $authorizationService,
+        OauthService $oauthService
+    ) {
+        $this->_klaviyoScopeSetting = $klaviyoScopeSetting;
+        $this->_messageManager = $messageManager;
```
remstone7, line 48: i don' think this is used, may want to remove it

Helper/ScopeSetting.php
```diff
@@ -20,6 +20,8 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     const KLAVIYO_PASSWORD = 'klaviyo_reclaim_user/klaviyo_user/password';
     const KLAVIYO_EMAIL = 'klaviyo_reclaim_user/klaviyo_user/email';
 
+    const KLAVIYO_OAUTH_NAME = 'klaviyo_reclaim_oauth/klaviyo_oauth/name';
```
remstone7, line 23: not required, but maybe:
```suggestion
    const KLAVIYO_OAUTH_NAME = 'klaviyo_reclaim_oauth/klaviyo_oauth/oauth_name';
```

remstone7 (commented):
looks good, some comments

Helper/ScopeSetting.php
klaviyojad: I am not a fan as now you repeat oauth the 3rd time.  can we get a second opinion, @smoucka ??

Observer/KlaviyoOauthObserver.php
klaviyojad: Good point, I really dronely put it cuz I saw other Observers doing it but dont know if it has any value

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 - Added a.fast.klaviyo.com to img-src csp whitelist
+- Oauth observer to create Magento2 Integration Oauth configuration
```
remstone7, line 11:
```suggestion
- OAuth observer to create Magento2 Integration OAuth configuration
```

Helper/ScopeSetting.php
remstone7: maybe integration_name ? again not required but figured we can try and be descriptive if we can

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Klaviyo\Reclaim\Observer;
+
+use Klaviyo\Reclaim\Helper\ScopeSetting;
+use Magento\Framework\Event\ObserverInterface;
+use Magento\Framework\Event\Observer;
```
remstone7, line 7: want to alphabetize these?

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,88 @@
...
+{
+    /**
+     * Klaviyo scope setting helper
+     * @var \Klaviyo\Reclaim\Helper\ScopeSetting $klaviyoScopeSetting
+     */
+    protected $_klaviyoScopeSetting;
+
+
+    /**
+     * @var IntegrationFactory $integrationFactory
```
remstone7, line 24: small nit, i guess it should follow
```suggestion
     * @var  \Magento\Integration\Model\IntegrationFactory $integrationFactory
```

Helper/ScopeSetting.php
remstone7: or consumer_name?

Observer/KlaviyoOauthObserver.php
```diff
@@ -0,0 +1,88 @@
...
+            // Code to create consumer
+            $consumer = $this->_oauthService->createConsumer(['name' => $consumerName]);
+            $consumerId = $consumer->getId();
+            $integration->setConsumerId($consumerId);
+            $integration->save();
+
+            // Code to grant permission
+            $this->_authorizationService->grantAllPermissions($integrationId);
+            
+        }catch(Exception $e){
```
remstone7, line 83:
```suggestion
        } catch(Exception $e) {
```

Helper/ScopeSetting.php
njparadis: IMO it's fine as is - you can see we follow a similar pattern for things like `klaviyo_reclaim_user/klaviyo_user/email`

Observer/KlaviyoOauthObserver.php
klaviyojad: is that a standard or just to be consistent?  I am not seeing that pattern for the other Observers

remstone7: consistent

klaviyojad: consistent means I shouldn't have the fully qualified names since it isnt done like that for the other Observers

remstone7: Yes, the other observers haven't been edited in a while, we should convert them over, its better to provide the full path since its easier to distinguish where the protected variable is coming from rather than having to look in other places in the file

remstone7 (approved):
lgtm, wait on merging until app side work is done

etc/adminhtml/system.xml
```diff
@@ -69,6 +82,7 @@
                 <field id="email" translate="label comment" type="text" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
                     <label>Email</label>
                 </field>
+                <comment>NOTE: Support for Klaviyo user authentication is set to deprecate.  We advise setting up access via the Setup Oauth tab</comment>
```
remstone7, line 85: I think we can get rid of this or change the description here since we won't be using the klaviyo user tabe?

klaviyojad: yes I am planning on removing it, the latest push was just to rebase with master

remstone7 (approved):

klaviyojad:
@ihor-sviziev user/password-based integration will continue to work for customers that have set it up.  There will be updated instructions in Help Center for users to migrate to OAuth to be published later today.  Also the UI will change in the Klaviyo App.

Setup/InstallData.php
```diff
@@ -70,29 +44,5 @@ public function install(
         catch (\Magento\Framework\Exception\LocalizedException $ex) {
             $this->_state->setAreaCode(\Magento\Framework\App\Area::AREA_ADMINHTML);
         }
```
ihor-sviziev, line 46: Looks like this file doesn't do anything anymore. So it could be safely removed. Right?

PR #94, opened 2020-12-10
Fix reclaim/checkout/url POST upon email update for subdirectories

ikk0:
Hi, this code didn't work properly. Using the url.build function is much better to build an URL compared to using the approach previously in this code.


remstone7 (approved):
good call on this 👍

njparadis:
Probably should make a similar update to: https://github.com/klaviyo/magento2-klaviyo/blob/master/view/frontend/templates/checkout/cart.phtml

njparadis:
Closing this out as this fix was included with #137 !

PR #100, opened 2021-03-10
sms consent at checkout

remstone7:


Setup/UpgradeSchema.php
```diff
@@ -0,0 +1,53 @@
...
+use Magento\Framework\DB\Ddl\Table;
+use Magento\Quote\Setup\QuoteSetup;
+use Magento\Sales\Setup\SalesSetup;
+
+class UpgradeSchema implements UpgradeSchemaInterface
+{
+
+    public function upgrade(SchemaSetupInterface $setup, ModuleContextInterface $context)
+    {
+//        if (version_compare($context->getVersion(), '2.0.0', '<')) {
```
remstone7, line 22: need to actually uncomment this for new version release

view/frontend/web/js/model/shipping-save-processor/default.js
```diff
@@ -0,0 +1,73 @@
...
+                    shipping_carrier_code: quote.shippingMethod().carrier_code,
+                    extension_attributes: {
+                        kl_sms_consent: kl_sms_consent,
+                        kl_email_consent: kl_email_consent,
+                    }
+                }
+            };
+
+            //TODO Remove
+            console.log('debug sms', kl_sms_consent);
```
remstone7, line 51: will remove these

etc/events.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<config xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Event/etc/events.xsd">
+    <event name="sales_model_service_quote_submit_before">
```
remstone7, line 4: should i do this post order instead of submit? probably?

Plugin/CheckoutLayoutPlugin.php
```diff
@@ -0,0 +1,76 @@
...
+                'sortOrder' => $this->_klaviyoScopeSetting->getConsentAtCheckoutSMSConsentSortOrder(),
+                'id' => 'kl_sms_consent',
+            ];
+
+            $jsLayout['components']['checkout']['children']['steps']['children']['shipping-step']['children']['shippingAddress']['children']['shipping-address-fieldset']['children']['kl_sms_consent'] = $smsConsentCheckbox;
+        }
+        // Open to ideas here, since we don't overwrite the customer-email section
+        // we need to distinguish if the customer is logged in or not, object manager is an easy way to do so
+        $objectManager = \Magento\Framework\App\ObjectManager::getInstance();
+        $customerSession = $objectManager->get('Magento\Customer\Model\Session');
```
remstone7, line 46: we don't have access to the customer session, this is frowned upon but no other way to see if a user is logged in or not.

view/frontend/web/js/model/shipping-save-processor/default.js
```diff
@@ -0,0 +1,73 @@
...
+    errorProcessor,
+    fullScreenLoader,
+    selectBillingAddressAction
+) {
+    'use strict';
+
+    return {
+        saveShippingInformation: function() {
+            var payload;
+            var kl_sms_consent = $('[name="custom_attributes[kl_sms_consent]"]').is(':checked');
```
remstone7, line 30: these are the custom attributes we added to the shipping form

klaviyojad:
I am not sure if its just my setup but it seems like the Admin Settings should show a drop down but doesnt?
image

klaviyojad:
^ sorry about the above. It was due to not having set the public and private key in the general tab

Plugin/CheckoutLayoutPlugin.php
```diff
@@ -0,0 +1,76 @@
...
+                    'id' => 'kl_email_consent',
+                ],
+                'dataScope' => 'shippingAddress.custom_attributes.kl_email_consent',
+                'label' => 'Sign Up for Email Newsletters',
+                'description' => $this->_klaviyoScopeSetting->getConsentAtCheckoutEmailConsentText(),
+                'provider' => 'checkoutProvider',
+                'visible' => true,
+                'checked' => false,
+                'validation' => [],
+                'sortOrder' => 0,
```
klaviyojad, line 66: im curious how the email consent doesn't need a sortOrder setting and yet appears in the right place for a checkout yet sms consent takes default value of 200 and is configurable and seems to work

Observer/SaveOrderMarketingConsent.php
```diff
@@ -0,0 +1,81 @@
...
+
+        if (
+            $this->_klaviyoScopeSetting->getWebhookSecret()
+            && $this->_klaviyoScopeSetting->getConsentAtCheckoutSMSIsActive()
+        ) {
+            $data = array(
+                "data" => array(
+                    array(
+                        "customer" => array(
+                            "email" => $quote->getCustomerEmail(),
```
klaviyojad, line 64: i guess you want to add all the other stuff like first name, last name and other pieces of address if possible?

Setup/InstallSchema.php
```diff
@@ -0,0 +1,51 @@
...
+    {
+        $installer = $setup;
+        $installer->startSetup();
+
+        $eavTable1 = $installer->getTable('quote');
+        $eavTable2 = $installer->getTable('sales_order');
+
+        $columns = [
+            'kl_sms_consent' => [
+                'type' => \Magento\Framework\DB\Ddl\Table::TYPE_TEXT,
```
klaviyojad, line 31: Unless I am mistaken that these values are not just true/false .  why are you using text type instead of smallint(5) teh convention used in Magento MySql for everything else representing a boolean?

image

view/frontend/web/js/model/shipping-save-processor/default.js
```diff
@@ -0,0 +1,73 @@
...
+            console.log('debug email', kl_email_consent);
+
+            fullScreenLoader.startLoader();
+
+            return storage.post(
+                resourceUrlManager.getUrlForSetShippingInformation(quote),
+                JSON.stringify(payload)
+            ).done(
+                function(response) {
+                    quote.setTotals(response.totals);
```
klaviyojad, line 61: why are we doing all the stuff here?  Is this replacing the existing functionality?  what if the customer has other extensions that run affecting what happens here i.e if they had a 3d party coupon processor

Observer/SaveOrderMarketingConsent.php
remstone7: we shouldn't need to here, that is all done through app side code, this is really just for list subscriptions, if someone hits this stage, all the other customer properties will come over as well

Plugin/CheckoutLayoutPlugin.php
remstone7: great question, i thought about adding it, it's basically because the email form will always be at the top of the page, this is inside the shipping address form which will be rendered after the email form, so putting it at top makes sense.  For the phone number, 200 is just arbitrary to make sure it's at the bottom, but customers CAN customize the sort order of the shipping fields and if they did, phone number could be moved so we should allow them to move the consent box as well

Setup/InstallSchema.php
remstone7: totally, should be a bool 👍

view/frontend/web/js/model/shipping-save-processor/default.js
remstone7: yea this is basically magento standard for doing this based off all documentation and what similar apps do.

Setup/InstallSchema.php
remstone7: leaving as a textfield due to tiny int's on different mysql versions, it also allows us to potentially add more in the future?

CHANGELOG.md
```diff
@@ -6,8 +6,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [Unreleased]
 
-### Fixed
+#### Added
+- SMS Consent at checkout
+- Email consent at checkout
+- Concent at checkout admin tab
```
jordanallain, line 12: Consent*

Plugin/CheckoutLayoutPlugin.php
jordanallain: side note, would we be able to utilize this for recording started checkouts for logged in users?

remstone7: we already handle that in the email.js file

Observer/SaveOrderMarketingConsent.php
```diff
@@ -0,0 +1,98 @@
...
+        $webhookSecret = $this->_klaviyoScopeSetting->getWebhookSecret();
+        $updatedAt = $quote->getUpdatedAt();
+
+        $data = array("data" => array());
+
+        if (
+            $webhookSecret
+            && $this->_klaviyoScopeSetting->getConsentAtCheckoutSMSIsActive()
+        ) {
+            $data["data"][] = array(
```
jordanallain, line 64: are the extra empty brackets so that we nest the data under `0` ?

etc/config.xml
```diff
@@ -17,5 +17,30 @@
...
             </klaviyo_webhooks>
         </klaviyo_reclaim_webhook>
+        <klaviyo_reclaim_consent_at_checkout>
+            <email_consent>
+                <consent_text>Subscribe for email updates</consent_text>
+            </email_consent>
+        </klaviyo_reclaim_consent_at_checkout>
+        <klaviyo_reclaim_consent_at_checkout>
+            <sms_consent>
+                <consent_text>**By checking this box and entering your phone number above, you consent to receive marketing text messages at the number provided from {company name], including messages sent by autodialer. Consent is not a condition of any purchase. Message and data rates may apply. Message frequency varies. Reply HELP for help or STOP to cancel. View our Privacy Policy and Terms of Service.</consent_text>
```
jordanallain, line 27: Small nit, should this placeholder use the same opening and closing characters? `{company name]`

remstone7: nice catch 👍

Observer/SaveOrderMarketingConsent.php
remstone7: just how to append to an array

jordanallain (approved):

cpcurtis1218 (approved):

klaviyojad (approved):

PR #106, opened 2021-05-03
Use store ids in the ProductDeleteBefore Observer

jordanallain:
Fixes bug to use store_ids instead of website_ids for delete product webhook.

Also fixes bug where we were looking for versions greater than 2.0.0 in the UpgradeSchema.

### DIFF
--- a/Observer/ProductDeleteBefore.php
+++ b/Observer/ProductDeleteBefore.php
@@ -54,8 +54,8 @@ class ProductDeleteBefore implements ObserverInterface
     public function execute(Observer $observer)
     {
         $product = $observer->getEvent()->getProduct();
-        $websiteIds = $product->getWebsiteIds();
-        $storeIdKlaviyoMap = $this->_klaviyoScopeSetting->getStoreIdKlaviyoAccountSetMap($websiteIds);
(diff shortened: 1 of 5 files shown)


Observer/ProductDeleteBefore.php
```diff
@@ -54,8 +54,8 @@ public function __construct(
     public function execute(Observer $observer)
     {
         $product = $observer->getEvent()->getProduct();
-        $websiteIds = $product->getWebsiteIds();
-        $storeIdKlaviyoMap = $this->_klaviyoScopeSetting->getStoreIdKlaviyoAccountSetMap($websiteIds);
+        $storeIds = $product->getStoreIds();
```
klaviyojad, line 57: I could see downstream the storeIds is set inside the $data array.  Why wasnt it explicitly declared before? Am I missing something?

remstone7: this is a bug, we should have been getting storeIds of the product not websiteIds

remstone7 (approved):

remstone7 (approved):

remstone7 (approved):

PR #107, opened 2021-05-18
Update checkout rebuilding

remstone7:


smoucka (commented):
Should we just bump the version in this pr as well?

CHANGELOG.md
```diff
@@ -6,9 +6,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [Unreleased]
 
+### [2.2.0] - 2021-08-18
```
njparadis, line 9: I think this should be 2021-05-18

Plugin/Api/CartSearchRepository.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
+
+namespace Klaviyo\Reclaim\Plugin\Api;
+
+
+use Magento\Quote\Api\CartRepositoryInterface;
+use Magento\Quote\Api\Data\CartExtensionFactory;
+use Magento\Quote\Model\QuoteIdToMaskedQuoteId;
```
klaviyojad, line 9: what happens to customers on version pre 2.3?  If you look at https://github.com/magento/magento2/tree/2.2/app/code/Magento/Quote/Model it does not exist.

remstone7: we'll probably have to add this as a dependency

njparadis (approved):

webdev2k21:
Hi,

On Magento 2.3.3 when we apply coupon code getting the error as below.

{"messages":{"error":[{"code":500,"message":"Fatal Error: 'Uncaught TypeError: Argument 2 passed to Klaviyo\\Reclaim\\Plugin\\Api\\CartSearchRepository::afterGetList() must be an instance of Magento\\Quote\\Model\\CartSearchResults, instance of Magento\\Framework\\Api\\SearchResults given, defined in \/var\/www\/html\/magento2\/app\/code\/Klaviyo\/Reclaim\/Plugin\/Api\/CartSearchRepository.php:49

remstone7:
Thanks @webdev2k21 we will look into this

klaviyojad:
@webdev2k21 it appears to be a problem where your version does not have it like the 2.3 version magento has here: https://github.com/magento/magento2/blob/2.3/app/code/Magento/Quote/Model/CartSearchResults.php
I know they must have updated it after the fact because I can see 2.3.3 does not have it https://github.com/magento/magento2/tree/2.3.3/app/code/Magento/Quote/Model

I wonder if there is way to be still on 2.3 and get the latest updates for module-quote ?

Plugin/Api/CartSearchRepository.php
```diff
@@ -0,0 +1,76 @@
...
+            $quote->setExtensionAttributes($extensionAttributes);
+        }
+
+        return $searchResult;
+    }
+
+    public function getMaskedIdFromQuoteId($quoteId)
+    {
+        try {
+            $quoteId = $this->quoteIdMask->execute($quoteId);
```
ihor-sviziev, line 70: @remstone7 @klaviyojad 
FYI this change causing ~2x slow down on the quotes list retrieval request, but that's an issue in the Magneto core.
More details here: https://github.com/magento/magento2/pull/33324

PR #108, opened 2021-05-24
m2 checkout for under 2.3

remstone7:


smoucka (commented):
Should this just be incorporated into version 3?

remstone7:
> Should this just be incorporated into version 3?

no because version 3 will only support magento version 2.3 + this is the "last" patch to support versions 2.2 and under

smoucka (approved):

jordanallain (approved):

PR #115, opened 2021-06-03
Added to Cart m2 initial commit

siddwarkhedkar:
In addition to the method of capturing the event outlined in the RFC, I am also creating a variable in the checkout session to be used after the Sales Quote is save. The reason for this >> the quote is empty and not created when the cart is empty, making it difficult to send over the MaskedQuoteId to Klaviyo. 

Creating a session variable and using this after the sales quote is saved. The `Sales_quote_save_after` event has an observer which will make use of this data and then make a track request to Klaviyo


Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+            $cartItemCategories = $this->uniqueArrayOfStrings( $cartItemCategories, $itemCategories );
+        }
+
+        return array(
+            '$value' => (float) $quote->getBaseGrandTotal() + $addedItem->getPrice(),
+            'ItemNames' => (array) $cartItemNames,
+            'Items' => (array) $items,
+            'ItemCount' => (int) $cartQty,
+            'Categories' => (array) $cartItemCategories,
+            '$service' => 'magento_two'
```
remstone7, line 128: Please remove the service, this isn't something we do for any other integrations
```suggestion

```

Observer/SalesQuoteSaveAfter.php
```diff
@@ -0,0 +1,92 @@
...
+    )
+    {
+        $this->_dataHelper = $dataHelper;
+        $this->_scopeSetting = $scopeSetting;
+        $this->_cartSearchRepository = $cartSearchRepository;
+        $this->_checkoutsession = $checkoutsession;
+    }
+
+    public function execute( Observer $observer )
+    {
```
remstone7, line 61: is there a reason we can't just package the data in this observer rather than needing two different ones? it seems like we should have everything needed in this observer to fire it? This has a quote, which allows us to get a masked id. it also provides us with the data to get `getLastAddedItem` from the quote and all of the items that exist in the cart?

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Helper function to add items from cart to Added to Cart payload
+     * @param $quote
+     * @param $addedItem
+     * @return array
+     */
+    public function klBuildCartData( $quote, $addedItem )
+    {
+        $cartItems = $quote->getItems() ? $quote->getitems() : array();
```
klaviyojad, line 95: can you elaborate on the logic here?

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+        foreach ( $categoryIds as $id ) {
+            $category = $this->_categoryFactory->create()->load( $id );
+            $categoryName = $category->getName();
+            array_push( $categoryNames, $categoryName );
+        }
+
+        return $categoryNames;
+    }
+
+    public function uniqueArrayOfStrings( $array_one, $array_two ): array
```
klaviyojad, line 149: did you consider a Set Data structure as highlighted here: https://stackoverflow.com/questions/3764800/does-php-have-a-set-data-structure

Perhaps emulating a Set through the HashMap interface by setting your elements as an array key?

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+        $cartItems = $quote->getItems() ? $quote->getitems() : array();
+        $cartQty = 0;
+        $items = array();
+        $cartItemNames = array();
+        $cartItemCategories = array();
+
+        foreach( $cartItems as $item ) {
+            $product = $item->getProduct();
+            $cartItemId = $product->getId();
+            $itemCategories = $this->getCategoryName( $product->getCategoryIds() );
```
smoucka, line 104: Nit: the additional spacing is a Wordpress specific style convention. We don't need those here
```suggestion
            $itemCategories = $this->getCategoryName($product->getCategoryIds());
```

Observer/SalesQuoteProductAddAfter.php
smoucka: `getItems` can return null in which case we want to instead return an empty array. you can use a null coalescing operator (php 7+) here instead if you want to clean it up a bit further
```suggestion
        $cartItems = $quote->getItems() ?? array();
```

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+            $cartQty += $item->getQty();
+            array_push( $items, $currentProduct );
+            array_push( $cartItemNames, $itemName );
+            $cartItemCategories = $this->uniqueArrayOfStrings( $cartItemCategories, $itemCategories );
+        }
+
+        return array(
+            '$value' => (float) $quote->getBaseGrandTotal() + $addedItem->getPrice(),
+            'ItemNames' => (array) $cartItemNames,
+            'Items' => (array) $items,
```
smoucka, line 125: I think the event structure should probably mimic the Checkout Started event so that it's quick and easy to repurpose a flow template for an Abandoned Cart flow triggered ont he Added to Cart event. The array of cart line items should be under `event.Extra.LineItems`

Observer/SalesQuoteProductAddAfter.php
jordanallain: looks like it'll set `$cartItems` to what `getItems()` returns, or an empty array if `getItems()` doesn't return anything

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,157 @@
...
+    /**
+     * Retrieves category names from category IDs
+     * @param array $categoryIds
+     * @return array
+     */
+    public function getCategoryName( array $categoryIds )
+    {
+        $categoryNames = [];
+        foreach ( $categoryIds as $id ) {
+            $category = $this->_categoryFactory->create()->load( $id );
```
jordanallain, line 141: I don't think you need to create a new categoryFactory each time through the loop, should be able to create one before the loop and use that for each iteration.

Observer/SalesQuoteSaveAfter.php
siddwarkhedkar: addressed this in the message_queues PR, the quoteId doesn't get set in the SalesQuoteProductSaveAfter event which requires the other observer. 

The other observer will save the quote Id and one of the cron jobs will get the masked Quote ID before moving to the sync table.

siddwarkhedkar:
Closing in favor of #142

PR #119, opened 2021-06-07
Add missing store id variable

ihor-sviziev:
This variable used in class, but it's not actually defined. This PR fixes this.


ihor-sviziev:
That's not a good option. We should get current store id in the runtime, as the current store might be changed for instance in the Cron job.

PR #120, opened 2021-06-07
Add an ability to pass the Store ID during track event

ihor-sviziev:
Use case - we would like to send some events to Klaviyo in the cron job for a specific store (different stores synced to different Klaviyo accounts).
Right now, we can't specify to which Klavityo account we should send this event.

This PR brings an ability to send the event to a specific Klaviyo account (specified on store level).


ihor-sviziev:
@klaviyojad @remstone7, could you review this one as well?

remstone7 (approved):

PR #123, opened 2021-06-08
Keep existing extension attributes when extending shipping payload request

lsimeonov:
Currently when using the extension with another module there is a potential problem. 
When another module is also making changes to `extension_attributes` payload it gets overridden by Klaviyo. 

Magento made an effort on that part and gave possibility to extend the payload by utilising mixins. This is after version 2.2.2
I replaced the current shipping-save-processor override with new mixin approach on the mage/storage so it will work prior version 2.2.2.

And implemented the new way of extending the shipping payload on checkout. 

I was unable to find good solution to feature switch based on magento version.

Any ideas and comments are quite welcome.


view/frontend/requirejs-config.js
```diff
@@ -2,7 +2,17 @@ var config = {
     map: {
         '*': {
             KlaviyoCustomerData: 'Klaviyo_Reclaim/js/customer',
-            'Magento_Checkout/js/model/shipping-save-processor/default': 'Klaviyo_Reclaim/js/model/shipping-save-processor/default'
+        }
+    },
+    config: {
+        mixins: {
+            // Compatibility with Magento < 2.2.2
+            'mage/storage': {
```
remstone7, line 10: this is a nice solution for below 2.2.2 - i don't think it'll be required because the current version 3.0.5 + will only support 2.3.4+

lsimeonov: Thanks for the insights. I removed the 2.2.2 compatibility

view/frontend/web/js/mixin/shipping-payload-extender-mixin.js
```diff
@@ -0,0 +1,22 @@
...
+    'use strict';
+
+    /**
+     * This file works on Magento >= 2.2.2 only.
+     *
+     * @param  {Object} target
+     * @return {Object}
+     */
+    return function (target) {
+        return wrapper.wrap(target, function (o, payload) {
```
jordanallain, line 14: what is 'o' here?

jordanallain (approved):
Looks good, assuming it was tested.

PR #125, opened 2021-06-08
copied over changes from ihor's PR

klaviyojad:
Use case - we would like to send some events to Klaviyo in the cron job for a specific store (different stores synced to different Klaviyo accounts).
Right now, we can't specify to which Klavityo account we should send this event.

This PR brings an ability to send the event to a specific Klaviyo account (specified on store level).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [Unreleased]
 
+#### Added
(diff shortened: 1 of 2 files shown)


klaviyojad:
@ihor-sviziev I had to create a new PR since I could not rebase right from the other one you created after I added CHANGELOG changes

ihor-sviziev:
Hi @klaviyojad,
I rebased my branch and force pushed all changes in https://github.com/klaviyo/magento2-klaviyo/pull/120, so I think you can just merge it as is now

PR #129, opened 2021-07-16
Handle newsletter subscriptions in all areas

ihor-sviziev:
Pre-history:  we have a custom Checkout extension that has built-in functionality for subscribing to the newsletter. This subscription happening on the same request as "place order". This request usually happening in the `webapi_rest` area. As a result, in Magento, we have subscribers, but in Klaviyo, this information isn't getting sent, so **we're losing subscribers**.

My Pull Request changing handling [all possible areas](https://devdocs.magento.com/guides/v2.4/architecture/archi_perspectives/components/modules/mod_and_areas.html), including `webapi_rest`, `webapi_soap`, `graphql` and `crontab`,  as well as existing `frontend` and `adminhtml`.

### DIFF
--- a/etc/adminhtml/events.xml
+++ b/etc/adminhtml/events.xml
@@ -2,9 +2,6 @@
     <event name="config_data_save_before">
         <observer name="klaviyo_reclaim" instance="Klaviyo\Reclaim\Observer\PrivateApiKeyObserver" />
     </event>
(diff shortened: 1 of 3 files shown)


ihor-sviziev:
@klaviyojad @remstone7, could you review this Pull Request?

remstone7 (approved):

PR #131, opened 2021-08-27
newsletter fix

klaviyojad:
Fixing issue where anonymous users (not registered accounts) try to subscribe to newsletter and do not get synched to the associated list in Klaviyo


smoucka (approved):
lgtm

Issue #133, opened 2021-08-28
\Klaviyo\Reclaim\Plugin\Api\CartSearchRepository::afterGetList Infinite Loop


pmclain:
### Steps to reproduce

All cart interactions below should be done via API

1. Create cart
2. Add at least one product to cart
3. Apply coupon code to cart (must be a code)
4. Edit the cart product in the admin panel
5. Reload cart

### Expected Result

`/rest/{{store}}/V1/carts/mine` should return your cart ID

### Actual Result

An infinite loop is triggered when a quote has `quote.coupon_code <> NULL AND quote.trigger_recollect = 1`

Summarizing the call stack triggered by this combination:
https://github.com/klaviyo/magento2-klaviyo/blob/4e63ccd99ce74d6f72839fcccf764553596a0d45/Plugin/Api/CartSearchRepository.php#L57

https://github.com/magento/magento2/blob/ab815b51dad2a89b120cc53b26eefaee80a336f3/app/code/Magento/Quote/Model/QuoteIdToMaskedQuoteId.php#L53

https://github.com/magento/magento2/blob/ab815b51dad2a89b120cc53b26eefaee80a336f3/app/code/Magento/Quote/Model/QuoteRepository.php#L136

https://github.com/magento/magento2/blob/ab815b51dad2a89b120cc53b26eefaee80a336f3/app/code/Magento/Quote/Model/Quote.php#L2461

☝️ Recollection is triggered here because editing a product in the quote sets `quote.trigger_recollect = 1`

During recollection https://github.com/magento/magento2/blob/ab815b51dad2a89b120cc53b26eefaee80a336f3/app/code/Magento/SalesRule/Observer/CouponCodeValidation.php#L66
validates the coupon code. The validation calls `CartRepositoryInterface::getList` which re-triggers your plugin.

```php
$found = $this->cartRepository->getList(
  $this->criteriaBuilder->addFilter('main_table.' . CartInterface::KEY_ENTITY_ID, $quote->getId())
    ->create()
)->getItems();
```

At this point the request is infinitely looping until it timeouts or exhausts memory.

klaviyojad:
@pmclain I am seeing that this issue has been logged by @ihor-sviziev [here](https://github.com/magento/magento2/issues/33675) which leads me to think its a magento issue.  I will check if the latest version 2.4.3 it is fixed

ihor-sviziev:
@pmclain, please apply patch from the following PR:
[...]

PR #136, opened 2021-09-02
adding code changes to handle infinite loop issue

klaviyojad:
This pertains to github issue [here](https://github.com/klaviyo/magento2-klaviyo/issues/133) and implements some of the logic [here](https://github.com/magento/magento2/pull/33324) as its still unreleased code.  

It addresses the bug where if a quote with a coupon has `trigger_recollect = 1` then any api calls to retrieve the quote will hang forever. Instead of relying on the magento libraries to retrieve the masked quote id, we implement code to retrieve it ourselves from the DB

### DIFF
--- a/Model/Quote/QuoteIdMask.php
+++ b/Model/Quote/QuoteIdMask.php
@@ -1,0 +1,46 @@
+<?php
+/**
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Klaviyo\Reclaim\Model\Quote;
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
+/**
+ * QuoteIdMask Resource model
+ * @codeCoverageIgnore
+ */
+class QuoteIdMask extends AbstractDb
+{
+    /**
+     * Main table and field initialization
+     *
+     * @return void
+     */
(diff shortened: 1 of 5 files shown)


jordanallain:
Looks good, were you able to test it?

klaviyojad:
@jordanallain was tested by api call [store url]/rest/V1/carts/[quote id] that hangs for quote with coupon and trigger_recollect = 1 before change and doesnt after this change is made

jordanallain (approved):

Model/Quote/QuoteIdMask.php
```diff
@@ -0,0 +1,46 @@
...
+ * Copyright © Magento, Inc. All rights reserved.
+ * See COPYING.txt for license details.
+ */
+namespace Klaviyo\Reclaim\Model\Quote;
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
+/**
+ * QuoteIdMask Resource model
+ * @codeCoverageIgnore
+ */
+class QuoteIdMask extends AbstractDb
```
ihor-sviziev, line 12: FYI The resource models in Magento2 projects are usually located in the `Model/ResourceModel/` folder.
It's better to move it eventually to prevent confusion.

PR #137, opened 2021-09-03
Use Magento URL builder to build URLs

njparadis:
Implements change suggested in #94 and also implements this URL builder on cart.phtml template

Also replaces reference to deprecated _learnq functionality (`_learnq.identify.$email`) with an equivalent (`_learnq.isIdentified()`)

### DIFF
--- a/view/frontend/templates/checkout/cart.phtml
+++ b/view/frontend/templates/checkout/cart.phtml
@@ -5,7 +5,7 @@
            'mage/url'
         ], function (jQuery, url) {
           jQuery.ajax({
-              url: url.build('/reclaim/checkout/reload'),
+              url: url.build('reclaim/checkout/reload'),
               method: 'POST',
               data: {}
           });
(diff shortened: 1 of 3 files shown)


jordanallain:
assuming this was tested, lgtm! tiny nit, you could alphabetize the library names that are being required.

jordanallain (approved):

smoucka (approved):

view/frontend/templates/checkout/cart.phtml
```diff
@@ -1,10 +1,11 @@
...
     <script>
         require([
-           'jquery'
-        ], function (jQuery) {
+           'jquery',
+           'mage/url'
+        ], function (jQuery, url) {
           jQuery.ajax({
-              url: window.location.protocol + '//' + window.location.hostname + '/reclaim/checkout/reload',
+              url: url.build('/reclaim/checkout/reload'),
```
remstone7, line 8: does url.build not add the `/` ? it may handle it for both cases just want to make sure since we use it with and w/out a `/` as a prefix on the url

njparadis: Tested this out - seems url.build is smart enough to put the slash in between the base URL and the provided route if it's not there already

PR #138, opened 2021-09-20
Fix to Expose SMS Consent For Logged In Users with Default Address Set

klaviyojad:
Problem Statement:
Usually When an user adds a product to the shopping cart and proceeds with the checkout process.  The checkbox for SMS consent would appear just below the field for Phone Number as illustrated below:

image2

If the user happens to be logged in and has default address set, user is not given the opportunity to enter address details like we saw for the first example.  Instead you see a uneditable box with the address info as illustrated below:

image1

Solution:
Expose just between the default address and where user has to select the shipping method the redundant field for Phone Number with SMS consent checkbox underneath:

image3

### DIFF
--- a/Plugin/CheckoutLayoutPlugin.php
+++ b/Plugin/CheckoutLayoutPlugin.php
@@ -39,8 +39,8 @@ class CheckoutLayoutPlugin
 
     public function afterProcess(\Magento\Checkout\Block\Checkout\LayoutProcessor $processor, $jsLayout)
     {
-        if ($this->_klaviyoScopeSetting->getConsentAtCheckoutSMSIsActive()) {
-
+        if ($this->_klaviyoScopeSetting->getConsentAtCheckoutSMSIsActive())
+        {
             $smsConsentCheckbox = [
                 'component' => 'Magento_Ui/js/form/element/abstract',
                 'config' => [
--- a/Plugin/CheckoutLayoutPlugin.php
+++ b/Plugin/CheckoutLayoutPlugin.php
@@ -14,10 +14,36 @@ class CheckoutLayoutPlugin
         $this->_klaviyoScopeSetting = $klaviyoScopeSetting;
     }
 
+    /**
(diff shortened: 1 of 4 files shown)


Plugin/CheckoutLayoutPlugin.php
```diff
@@ -14,10 +14,36 @@ public function __construct(
...
+     * @return Magento\Customer\Model\Address|false
+     */
+    public function _getDefaultAddressIfSetForCustomer(
+        \Magento\Framework\App\ObjectManager $objectManager,
+        \Magento\Customer\Model\Session $customerSession
+    ) {
+        $address = false;
+        if ($customerSession->isLoggedIn()) {
+            $customerData = $customerSession->getCustomer()->getData();
+            $customerFactory = $objectManager->get('\Magento\Customer\Model\CustomerFactory')->create();
```
jordanallain, line 32: can you inject the CustomerFactory directly instead of injecting and using the ObjectManager to get it?

Plugin/CheckoutLayoutPlugin.php
```diff
@@ -38,14 +61,35 @@ public function afterProcess(\Magento\Checkout\Block\Checkout\LayoutProcessor $p
...
+                            'customScope' => 'shippingAddress',
+                            'template' => 'ui/form/field',
+                            'elementTmpl' => 'ui/form/element/input',
+                        ],
+                    'label' => 'Phone Number',
+                    'provider' => 'checkoutProvider',
+                    'sortOrder' => '120',
+                    'disabled' => true,
+                    'visible' => true,
+                    'value' => $address->getTelephone()
```
njparadis, line 84: Is it possible for a customer to have a saved address without a phone number?

klaviyojad: @njparadis no telephone is a required property when saving an address

njparadis (approved):

PR #140, opened 2021-09-22
Message queues db setup

siddwarkhedkar:
Starting 2.3, Magento decided to deprecate InstallSchema/UpgradeSchema, InstallData/UpgradeData scripts to using a db_schema.xml file and call it [declarative schema](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/declarative-schema/). 
If any of our customers are using a version older than 2.3.4, they can't upgrade M2 extensions to post 3.0.5, so we decided to go ahead with implementing this approach to add new tables. 

Magento suggests creating Models for each table to carry out CRUD operations on table. So each of the three tables have their own Model,ResourceModel and Collection classes.

The structure of such Models should be as follows:
```
extension_root
`-- Models
    |-- New_table.php
    `-- ResourceModel
        |-- New_table.php
        `-- New_table
            `-- Collection.php
```

### DIFF
--- a/Model/Events.php
+++ b/Model/Events.php
@@ -21,6 +21,5 @@ class Events extends AbstractModel
     protected function _construct()
     {
         $this->_init('Klaviyo\Reclaim\Model\ResourceModel\Events');
-        parent::_construct();
     }
 }
--- a/Model/Events.php
+++ b/Model/Events.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\AbstractModel;
+
+class Events extends AbstractModel
+{
+    protected function _construct()
+    {
+        $this->_init('Klaviyo\Reclaim\Model\Resourcemodel\Events');
+        parent::_construct();
+    }
+}
--- a/Model/Events.php
+++ b/Model/Events.php
@@ -4,11 +4,23 @@ namespace Klaviyo\Reclaim\Model;
 
 use Magento\Framework\Model\AbstractModel;
 
+/**
+ * Klaviyo Events Model is used to represent the kl_events table created using db_schema.xml file using declarative schema.
+ * https://devdocs.magento.com/guides/v2.4/extension-dev-guide/declarative-schema/
+ * The Model doesn't have the capability to query the DB directly but is required by the ResourceModel to create Magento DataObject instances
+ *
+ * The kl_events table will be used as an intermediary between the Observers and kl_syncs table
+ * This table records all user activity events to be sent over to the Klaviyo app using Track API
+ * eg. Added to Cart
+ */
 class Events extends AbstractModel
 {
+    /**
+     * Define resource model
+     */
     protected function _construct()
     {
-        $this->_init('Klaviyo\Reclaim\Model\Resourcemodel\Events');
+        $this->_init('Klaviyo\Reclaim\Model\ResourceModel\Events');
         parent::_construct();
     }
 }
--- a/Model/KlaviyoCollection.php
+++ b/Model/KlaviyoCollection.php
@@ -9,6 +9,12 @@ use Magento\Framework\Model\ResourceModel\Db\Collection\AbstractCollection;
  */
 abstract class KlaviyoCollection extends AbstractCollection
 {
+    /**
+     * Collection Helper method to get rows to be moved to sync table or send over to Klaviyo app
+     * Takes in status to construct the where query and returns 500 rows
+     * @param $status
+     * @return mixed
+     */
     public function getRowsForSync($status)
     {
         $syncCollection = $this->addFieldToFilter('status',$status)
@@ -18,6 +24,13 @@ abstract class KlaviyoCollection extends AbstractCollection
         return $syncCollection;
     }
 
+    /**
+     * Collection Helper method to get rows to be deleted.
+     * Fetches all rows older than 2 days and having status `SYNCED` for sync table, `MOVED` for topic tables
+     * and returns ids of these rows
+     * @param $status
+     * @return mixed
+     */
     public function getIdsToDelete($status)
     {
         $now = new \DateTime('now');
@@ -37,6 +50,12 @@ abstract class KlaviyoCollection extends AbstractCollection
         return $idsToDelete;
     }
 
+    /**
+     * Collection Helper method to update statuses of rows that have been moved or attempted sync to Klaviyo app
+     * Takes in array of ids and what status to update them to as arguments
+     * @param $ids
+     * @param $status
+     */
     public function updateRowStatus($ids, $status)
     {
         if (empty($ids)) {return;}
@@ -48,6 +67,10 @@ abstract class KlaviyoCollection extends AbstractCollection
         );
     }
 
+    /**
+     * Collection Helper method to delete rows that have been deemed fit for removal.
+     * @param $ids
+     */
     public function deleteRows($ids)
     {
         if (empty($ids)) {return;}
--- a/Model/KlaviyoCollection.php
+++ b/Model/KlaviyoCollection.php
@@ -17,8 +17,8 @@ abstract class KlaviyoCollection extends AbstractCollection
      */
     public function getRowsForSync($status)
     {
-        $syncCollection = $this->addFieldToFilter('status',$status)
-            ->addOrder('id',self::SORT_ORDER_ASC)
+        $syncCollection = $this->addFieldToFilter('status', $status)
+            ->addOrder('id', self::SORT_ORDER_ASC)
             ->setPageSize(500);
 
         return $syncCollection;
--- a/Model/KlaviyoCollection.php
+++ b/Model/KlaviyoCollection.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\ResourceModel\Db\Collection\AbstractCollection;
+
+/**
+ * Klaviyo abstract collection class used for inheritance
+ */
+abstract class KlaviyoCollection extends AbstractCollection
+{
+    public function getRowsForSync($status)
+    {
+        $syncCollection = $this->addFieldToFilter('status',$status)
+            ->addOrder('id',self::SORT_ORDER_ASC)
+            ->setPageSize(500);
+
+        return $syncCollection;
+    }
+
+    public function getIdsToDelete($status)
+    {
+        $now = new \DateTime('now');
+        $date = $now->sub(new \DateInterval('P2D'))
+            ->format('Y-m-d H:i:s');
+
+        $tableName = $this->getMainTable();
+
+        $idsToDelete = $this->getConnection()->fetchAll( "select id from (
+                                                            select id, timestampdiff(day, created_at, \"$date\") as row_age_in_days
+                                                            from $tableName
+                                                            where status = '".$status."'
+                                                            having row_age_in_days > 2
+                                                          ) as age_of_rows;"
+        );
+
+        return $idsToDelete;
+    }
+
+    public function updateRowStatus($ids, $status)
+    {
+        if (empty($ids)) {return;}
+
+        $this->getConnection()->update(
+            $this->getMainTable(),
+            $bind = ['status' => $status],
+            $where = ['id IN(?)' => $ids]
+        );
+    }
+
+    public function deleteRows($ids)
+    {
+        if (empty($ids)) {return;}
+
+        $this->getConnection()->delete(
+            $this->getMainTable(),
+            $where = ['id IN(?)' => $ids]
+        );
+    }
+}
--- a/Model/ResourceModel/Events.php
+++ b/Model/ResourceModel/Events.php
@@ -50,21 +50,4 @@ class Events extends AbstractDb
             $where
         );
     }
-
-    public function getIdsToDelete()
-    {
-        $now = new \DateTime('now');
-        $date = $now->sub( new \DateInterval('P2D') )->format('Y-m-d H:i:s');
-
-        $tableName = $this->getMainTable();
-
-        return $this->getConnection()->fetchAll( "select id from (
-                                                            select id, timestampdiff(day, created_at, \"$date\") as row_age_in_days
-                                                            from $tableName
-                                                            where status = '".self::MOVED."'
-                                                            having row_age_in_days > 2
-                                                          ) as age_of_rows;"
-        );
-
-    }
 }
--- a/Model/ResourceModel/Events.php
+++ b/Model/ResourceModel/Events.php
@@ -3,15 +3,16 @@
 namespace Klaviyo\Reclaim\Model\ResourceModel;
 
 use Klaviyo\Reclaim\Setup\SchemaInterface;
-use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
 
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
 /**
  * Klaviyo Events Table ResourceModel.
  *
  * The ResourceModel for any Model has the ability to query and make transactions with its associated table in the database.
- * This queries through the Zend\Db adapter which is connected to the Databse using the etc/env.php file.
+ * This queries through the Zend\Db adapter which is connected to the Database using the etc/env.php file.
  * The ResourceModel requires the Model to create DataObject instances and requires the table name and its idFieldName
  * to be defined.
+ * https://devdocs.magento.com/guides/v2.4/architecture/archi_perspectives/persist_layer.html
  */
 class Events extends AbstractDb
 {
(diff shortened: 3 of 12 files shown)


smoucka (commented):
Some general comments first.

[Would you mind adding some file-level doc blocks](https://pear.php.net/manual/en/standards.header.php)? Not sure if Magento has a specific style for this type of block but that link is for the PEAR standards. I think this could be really useful for folks who might work on some of this functionality in the future to easily understand how these various files make up the message queue feature. Much easier to have this documented in the code instead of the wiki or elsewhere. I'd love to see some elaboration around this sentence from the PR description and a link to the documentation where this is detailed.
>Magento suggests creating Models for each table to carry out CRUD operations on table. So each of the three tables have their own Model,ResourceModel and Collection classes.

Can you please add docstrings for the methods? Feel free to direct to the parent class methods if it makes sense.

I'm seeing a lot of extra spaces a la WordPress styling, can we make sure to clean that up so we're consistent with Magento styling?
https://github.com/klaviyo/magento2-klaviyo/pull/140/files#diff-05cdb3d45dbfc08c0c94488c55ec9cdb66b9a25e0279b42138e02dcaec1bcb32R18

Going to take a closer look at the logic and will review again this morning.

Model/Events.php
```diff
@@ -0,0 +1,14 @@
...
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\AbstractModel;
+
+class Events extends AbstractModel
+{
+    protected function _construct()
+    {
+        $this->_init('Klaviyo\Reclaim\Model\Resourcemodel\Events');
+        parent::_construct();
```
smoucka, line 12: Does this do anything? The parent class method is empty.

https://github.com/magento/magento2/blob/86dc050f2669617d1b34bc89a4a14522f008bf76/lib/internal/Magento/Framework/Model/AbstractModel.php#L204

Model/ResourceModel/Events/Collection.php
```diff
@@ -0,0 +1,48 @@
...
+        $idsToDelete = $this->getConnection()->fetchAll( "select id from (
+                                                            select id, timestampdiff(day, created_at, \"$date\") as row_age_in_days
+                                                            from $tableName
+                                                            where status = '".self::MOVED."'
+                                                            having row_age_in_days > 2
+                                                          ) as age_of_rows;"
+        );
+
+        return $idsToDelete;
+    }
```
smoucka, line 47: Knowing that we've got the cron logic coming in another PR, wondering if this shouldn't live there instead. I'm looking at some other cron cleanup logic e.g.
https://github.com/magento/magento2/blob/86dc050f2669617d1b34bc89a4a14522f008bf76/app/code/Magento/Sales/Model/CronJob/CleanExpiredOrders.php#L56

Do we want to be consistent and follow the pattern Magento uses and build that query in the cron class?

I could be way off base here in which case just say so.

Model/ResourceModel/Products.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model\ResourceModel;
+
+use Klaviyo\Reclaim\Helper\Logger;
+
+class Products extends \Magento\Framework\Model\ResourceModel\Db\AbstractDb
+{
+    protected $_klaviyoLogger;
```
smoucka, line 9:
```suggestion
    protected $_klaviyoLogger;

```

Model/ResourceModel/Products.php
```diff
@@ -0,0 +1,48 @@
...
+        $this->_init('kl_products', 'id');
+    }
+
+    public function __construct(
+        \Magento\Framework\Model\ResourceModel\Db\Context $context,
+        Logger $klaviyoLogger
+    )
+    {
+        parent::__construct($context);
+        $this->_klaviyoLogger = $klaviyoLogger;
```
smoucka, line 21: Are we using the logger anywhere here?

Model/ResourceModel/Products.php
```diff
@@ -0,0 +1,48 @@
...
+        if (empty($ids)) {
+            return;
+        }
+
+        $this->getConnection()->delete(
+            $this->getMainTable(),
+            ['id IN(?)' => $ids]
+        );
+    }
+}
```
smoucka, line 48: Looks like this logic is duplicated in the events model as well. Could be beneficial to create an abstract class for these methods especially if we intend to add other resource models in the future e.g. orders.

Model/ResourceModel/Syncs.php
```diff
@@ -0,0 +1,81 @@
...
+    protected function _construct()
+    {
+        $this->_init('kl_sync', 'id');
+    }
+
+    public function updateRowsToSynced($ids)
+    {
+        if (empty($ids)) {
+            return;
+        }
```
smoucka, line 21: This logic is being repeated all over the place. Is it necessary or is there a way to do this better? can we check in the callers maybe?

smoucka (commented):
Made some comments on certain resource models but I didn't want to double or triple up the same comments, so comments may apply to other models as well.

Model/Events.php
njparadis: I think it's also called by the parent `__construct()` function which this class inherits: https://github.com/magento/magento2/blob/86dc050f2669617d1b34bc89a4a14522f008bf76/lib/internal/Magento/Framework/Model/AbstractModel.php#L173

etc/db_schema.xml
```diff
@@ -0,0 +1,75 @@
+<?xml version="1.0"?>
+<schema xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:Setup/Declaration/Schema/etc/schema.xsd">
+    <table name="kl_sync" resource="default" engine="innodb" comment="Klaviyo Sync table">
+        <column xsi:type="int" name="id" padding="10" unsigned="true" nullable="false" identity="true"
+                comment="Primary key"/>
+        <column xsi:type="text" name="payload" nullable="true" comment="Event payload"/>
+        <column xsi:type="varchar" name="status" nullable="false" length="30" comment="Status of row (NEW, SYNCED...)"/>
+        <column xsi:type="varchar" name="topic" nullable="false" length="255" comment="Klaviyo Topic name"/>
+        <column xsi:type="text" name="klaviyo_id" nullable="true" length="100" comment="Klaviyo ID"/>
```
klaviyojad, line 9:
```suggestion
        <column xsi:type="text" name="klaviyo_id" nullable="true" comment="Klaviyo ID"/>
```
When you use a TEXT type for a column you don’t have to specify a storage length like for CHAR and VARCHAR because would hit issue [here](https://github.com/magento/devdocs/issues/7029)

etc/db_schema.xml
```diff
@@ -0,0 +1,75 @@
...
+            <column name="updated_at"/>
+        </index>
+    </table>
+    <table name="kl_products" resource="default" engine="innodb" comment="Klaviyo Products topic table">
+        <column xsi:type="int" name="id" padding="10" unsigned="true" nullable="false" identity="true"
+                comment="Primary key"/>
+        <column xsi:type="text" name="payload" nullable="true" comment="Event payload"/>
+        <column xsi:type="varchar" name="status" nullable="false" length="30" comment="Status of row (NEW, FAILED...)"/>
+        <column xsi:type="varchar" name="topic" nullable="false" length="255" comment="Klaviyo webhook topic"/>
+        <column xsi:type="text" name="klaviyo_id" nullable="true" length="100" comment="Klaviyo ID"/>
```
klaviyojad, line 34:
```suggestion
        <column xsi:type="text" name="klaviyo_id" nullable="true"  comment="Klaviyo ID"/>
```
same as above

Model/ResourceModel/Events/Collection.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    protected function _construct()
+    {
+        $this->_init( 'Klaviyo\Reclaim\Model\Events', 'Klaviyo\Reclaim\Model\ResourceModel\Events' );
+    }
+
+    public function getEventsToUpdate()
+    {
+        $eventsCollection = $this->addFieldToSelect( ['id','event','payload','user_properties'] )
+            ->addFieldToFilter( 'status','New' )
```
klaviyojad, line 24:
```suggestion
            ->addFieldToFilter( 'status','NEW' )
```
We really need to have a consistent naming convention.  In most places I see it in ALL Caps

etc/db_schema.xml
```diff
@@ -0,0 +1,75 @@
...
+            <column name="created_at"/>
+        </index>
+        <index referenceId="KL_PRODUCTS_UPDATED_AT">
+            <column name="updated_at"/>
+        </index>
+    </table>
+    <table name="kl_events" resource="default" engine="innodb" comment="Klaviyo Events topic table">
+        <column xsi:type="int" name="id" padding="10" unsigned="true" nullable="false" identity="true"
+                comment="Primary key"/>
+        <column xsi:type="varchar" name="status" nullable="true" length="30" comment="Status of row (New, Moved)"/>
```
klaviyojad, line 56:
```suggestion
        <column xsi:type="varchar" name="status" nullable="true" length="30" comment="Status of row (NEW, MOVED)"/>
```
to reflect the actual format of values saved

Model/Events.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\AbstractModel;
+
+class Events extends AbstractModel
```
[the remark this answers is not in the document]
siddwarkhedkar: Addressing this feedback

Model/ResourceModel/Events.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->_init(SchemaInterface::KL_EVENTS_TOPIC_TABLE, 'id');
+    }
+
+    public function __construct(
+        \Magento\Framework\Model\ResourceModel\Db\Context $context
+    )
+    {
+        parent::__construct($context);
+    }
```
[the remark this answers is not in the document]
siddwarkhedkar: wasn't necessary, removing this __construct method in feedback

Model/ResourceModel/Events/Collection.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model\ResourceModel\Events;
+
+use Magento\Framework\Model\ResourceModel\Db\Collection\AbstractCollection;
+
+class Collection extends AbstractCollection
+{
+    protected $_idFieldName = 'id';
```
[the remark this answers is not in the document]
siddwarkhedkar: cleaning up in feedback

Model/ResourceModel/Events/Collection.php
siddwarkhedkar: It is cleaner to have this live in a AbstractClass(KlaviyoCollection) which extends the Magento Abstract collection and all Klaviyo collections inherit methods from this AbstractClass, see feedback commit

Model/Events.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Klaviyo Events Model is used to represent the kl_events table created using db_schema.xml file using declarative schema.
+ * https://devdocs.magento.com/guides/v2.4/extension-dev-guide/declarative-schema/
+ * The Model doesn't have the capability to query the DB directly but is required by the ResourceModel to create Magento DataObject instances
+ *
+ * The kl_events table will be used as an intermediary between the Observers and kl_syncs table
+ * This table records all user activity events to be sent over to the Klaviyo app using Track API
+ * eg. Added to Cart
+ */
```
smoucka, line 15: This is fantastic, thanks for adding!

Model/Events.php
smoucka: Yep, so my question is do we need to call the parent method here (and in the other model classes)? If it's left blank so child classes implement it what is this call to the parent method doing?

Model/KlaviyoCollection.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Klaviyo abstract collection class used for inheritance
+ */
+abstract class KlaviyoCollection extends AbstractCollection
+{
+    public function getRowsForSync($status)
+    {
+        $syncCollection = $this->addFieldToFilter('status',$status)
+            ->addOrder('id',self::SORT_ORDER_ASC)
```
smoucka, line 15: nit:
```suggestion
        $syncCollection = $this->addFieldToFilter('status', $status)
            ->addOrder('id', self::SORT_ORDER_ASC)
```

Model/ResourceModel/Events.php
```diff
@@ -0,0 +1,25 @@
...
+
+use Klaviyo\Reclaim\Setup\SchemaInterface;
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
+
+/**
+ * Klaviyo Events Table ResourceModel.
+ *
+ * The ResourceModel for any Model has the ability to query and make transactions with its associated table in the database.
+ * This queries through the Zend\Db adapter which is connected to the Databse using the etc/env.php file.
+ * The ResourceModel requires the Model to create DataObject instances and requires the table name and its idFieldName
```
smoucka, line 13: Can you elaborate on this? I'm not explicitly seeing the connection to the Events Model here. Can you either explain  the relationship between the ResourceModel and the Model or link to Magento documentation on this topic? I'm asking so we can have this code be self-documenting such that someone in the future can look at these event classes, be able to understand exactly how they interact (and interact with the Magento core code) and then implement a new table.

Model/ResourceModel/Products.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model\ResourceModel;
+
+use Klaviyo\Reclaim\Setup\SchemaInterface;
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
```
smoucka, line 6:
```suggestion

use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
```

Model/ResourceModel/Syncs.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model\ResourceModel;
+
+use Klaviyo\Reclaim\Setup\SchemaInterface;
+use Magento\Framework\Model\ResourceModel\Db\AbstractDb;
+/*
```
smoucka, line 7:
```suggestion

/**
```

Setup/SchemaInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Klaviyo\Reclaim\Setup;
+
+interface SchemaInterface
```
smoucka, line 5: Can you add a short doc block for this interface?

Model/KlaviyoCollection.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\ResourceModel\Db\Collection\AbstractCollection;
+
+/**
+ * Klaviyo abstract collection class used for inheritance
+ */
+abstract class KlaviyoCollection extends AbstractCollection
+{
+    public function getRowsForSync($status)
```
smoucka, line 12: Can you please add dock blocks on these methods?

smoucka (commented):
This is looking really good. Can you please add doc blocks for methods? Specifically thinking about the KlaviyoCollection class.

I love the extra detail in the class level doc-blocks, super helpful. I mentioned this in a comment on one of the classes but I'd really love to see enough detail (or links to detail) for how these three classes relate and then more precisely how they interact with the Magento core code. It's not necessary to copy the same detail across each set of classes e.g. events, products, syncs, but you can just drop a @see and point to the one class that does contain that level of detail.

Model/Events.php
siddwarkhedkar: yeah we probably don't need it, I just tested this, works without calling the parent construct, my bad

Model/Events.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Klaviyo\Reclaim\Model;
+
+use Magento\Framework\Model\AbstractModel;
+
+/**
+ * Klaviyo Events Model is used to represent the kl_events table created using db_schema.xml file using declarative schema.
+ * https://devdocs.magento.com/guides/v2.4/extension-dev-guide/declarative-schema/
+ * The Model doesn't have the capability to query the DB directly but is required by the ResourceModel to create Magento DataObject instances
```
siddwarkhedkar, line 10: @smoucka This is where I've explained how the model and resourceModel are related, I could add a link to the persistence layer documentation Magento has but above there's not much listed in documentation

https://devdocs.magento.com/guides/v2.4/architecture/archi_perspectives/persist_layer.html

smoucka (approved):

PR #142, opened 2021-10-07
Message queues -  Added to Cart observers and Cron logic

siddwarkhedkar:
2 Observers are being used to record the Added to Cart events to the kl_events table
1. SalesQuoteProductAddAfter >> This is triggered by Magento when a product is added to a quote (cart) by a user and this captures multiple scenarios:
  a. Product added from it's page
  b. Product added from wishlist
  c. Re-Order of product

At this point, the SalesQuoteProductAddAfter observer will construct the payload for the event and creates a session variable

The SalesQuoteSaveAfter event is triggered when the Quote is saved and this enables us to get the QuoteId and send over with the payload to be able to reconstruct the cart later. This observer will read from the checkout session variable from the previous observer, attach the quote Id and then write to the kl_event table

The cron jobs require the crontab.xml, cron_groups.xml files for setup. We have created a new group in cron_groups.xml for all klaviyo_syncs
The crontab.xml file is where all cron jobs are listed and associated with different methods to be run

### DIFF
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -42,6 +42,12 @@ class EventsTopic
     protected $_categoryFactory;
 
     /**
+     * Category Map of ids to names
+     * @var array
+     */
+    protected $categoryMap = [];
+
+    /**
      * @param Logger $klaviyoLogger
      * @param CollectionFactory $eventsCollectionFactory
      * @param SyncsFactory $klSyncFactory
@@ -82,7 +88,7 @@ class EventsTopic
 
         // Capture all events that have been moved and add data to Sync table
         foreach ($eventsData as $event){
-            if ($event['topic'] == 'Added To Cart') {
+            if ($event['event'] == 'Added To Cart') {
                 $event['payload'] = json_encode($this->replaceQuoteIdAndCategoryIds($event['payload']));
             }
 
@@ -98,7 +104,7 @@ class EventsTopic
                 $sync->save();
                 array_push($idsMoved, $event['id']);
             } catch (\Exception $e) {
-                $this->_klaviyoLogger->log(sprintf("Unable to move row: %s", $e));
+                $this->_klaviyoLogger->log(sprintf("Unable to move row: %s", $e->getMessage()));
             }
         }
 
@@ -143,48 +149,42 @@ class EventsTopic
      */
     public function replaceCategoryIdsWithNames(array $payload): array
     {
-        $categoryMap = $this->getCategoryMap($payload['Categories']);
+        $this->updateCategoryMap($payload['Categories']);
 
         foreach ($payload['Items'] as &$item){
-            $item['Categories'] = $this->searchCategoryMapAndReturnNames($item['Categories'], $categoryMap);
+            $item['Categories'] = $this->searchCategoryMapAndReturnNames($item['Categories']);
         }
 
-        $payload['AddedItemCategories'] = $this->searchCategoryMapAndReturnNames($payload['AddedItemCategories'], $categoryMap);
-        $payload['Categories'] = array_values($categoryMap);
+        $payload['AddedItemCategories'] = $this->searchCategoryMapAndReturnNames($payload['AddedItemCategories']);
+        $payload['Categories'] = $this->searchCategoryMapAndReturnNames($payload['Categories']);
 
         return $payload;
     }
 
     /**
-     * Retrieve categoryNames from category factory using Ids
+     * Retrieve categoryNames using categoryIds
      * @param array $categoryIds
-     * @return array
      */
-    public function getCategoryMap(array $categoryIds): array
+    public function updateCategoryMap(array $categoryIds)
     {
         $categoryFactory = $this->_categoryFactory->create();
-        $categoryMap = [];
 
         foreach($categoryIds as $categoryId){
-            $categoryMap[$categoryId] = $categoryFactory->load($categoryId)->getName();
+            if (!in_array($categoryId, $this->categoryMap)){
+                $this->categoryMap[$categoryId] = $categoryFactory->load($categoryId)->getName();
+            }
         }
-
-        return $categoryMap;
     }
 
     /**
      * Return array of CategoryNames from CategoryMap using ids
      * @param array $categoryIds
-     * @param array $categoryMap
      * @return array
      */
-    public function searchCategoryMapAndReturnNames(array $categoryIds, array $categoryMap): array
+    public function searchCategoryMapAndReturnNames(array $categoryIds): array
     {
-        $categoryNames = [];
-        foreach ($categoryIds as $id){
-            array_push($categoryNames, $categoryMap[$id]);
-        }
-
-        return $categoryNames;
+        return array_values(
+            array_intersect_key($this->categoryMap, array_flip($categoryIds))
+        );
     }
 }
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -53,8 +53,7 @@ class EventsTopic
     }
 
     /**
-     * Cron method used by klaviyo_events_topic cron job
-     * Reads rows from kl_events table and writes to the kl_sync table
+     *
      */
     public function moveRowsToSync()
     {
@@ -78,7 +77,7 @@ class EventsTopic
                 'status' => 'NEW',
                 'topic' => $event['event'],
                 'user_properties' => $event['user_properties'],
-                'payload' => $this->addMaskedQuoteIdToEventPayload(json_decode($event['payload'], true))
+                'payload' => $this->replaceQuoteIdwithMaskedQuoteId($event['payload'])
             ]);
             try {
                 $sync->save();
@@ -92,10 +91,6 @@ class EventsTopic
         $eventsCollection->updateRowStatus($idsMoved, 'MOVED');
     }
 
-    /**
-     * Cron method used by the klaviyo_events_cleanup cron job
-     * Delete all rows marked as MOVED and older than 2 days in the kl_events table
-     */
     public function deleteMovedRows()
     {
         // Delete rows that have been moved to sync table
@@ -106,20 +101,19 @@ class EventsTopic
     }
 
     /**
-     * Helper method to get MaskedQuoteId from quote_id_mask table and replace QuoteId in payload
-     * This needs to be done here since MaskedQuoteId is unavailable when rows are recorded in kl_events table
-     * @param $payload
+     * Helper function to replace QuoteId in Added To Cart payload with Masked Quote Id
+     * @param array $payload
      * @return false|string
      */
-    public function addMaskedQuoteIdToEventPayload( $payload )
+    public function replaceQuoteIdwithMaskedQuoteId( array $payload )
     {
-        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(( $payload['QuoteId'] ));
-        unset($payload['QuoteId']);
+        $decoded_payload = json_decode($payload, true);
+        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(( $decoded_payload['QuoteId'] ));
+        unset($decoded_payload['QuoteId']);
 
         return $payload = json_encode(array_merge(
-            $payload,
-            array('MaskedQuoteId' => $maskedQuoteId)
+            $decoded_payload,
+            ['MaskedQuoteId' => $maskedQuoteId]
         ));
     }
-
 }
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -53,7 +53,7 @@ class EventsTopic
     }
 
     /**
-     *
+     * Gets 500 rows from the kl_events table and move to the kl_sync table
      */
     public function moveRowsToSync()
     {
@@ -91,6 +91,9 @@ class EventsTopic
         $eventsCollection->updateRowStatus($idsMoved, 'MOVED');
     }
 
+    /**
+     * Deletes rows moved to the kl_syncs table that are older than 2 days
+     */
     public function deleteMovedRows()
     {
         // Delete rows that have been moved to sync table
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -22,7 +22,7 @@ class EventsTopic
     protected $_quoteIdMaskResource;
 
     /**
-     * Klaviyo Events Collection Facd ..ctory
+     * Klaviyo Events Collection Factory
      * @var CollectionFactory
      */
     protected $_eventsCollectionFactory;
@@ -63,14 +63,14 @@ class EventsTopic
             ->addFieldToSelect(['id','event','payload','user_properties'])
             ->getData();
 
-        if (empty( $eventsData )){
+        if (empty($eventsData)){
             return;
         }
 
         $idsMoved = [];
 
         // Capture all events that have been moved and add data to Sync table
-        foreach ( $eventsData as $event ){
+        foreach ($eventsData as $event){
             //TODO: This can probably be done as one bulk update instead of individual inserts
             $sync = $this->_klSyncFactory->create();
             $sync->setData([
@@ -108,10 +108,10 @@ class EventsTopic
      * @param array $payload
      * @return false|string
      */
-    public function replaceQuoteIdwithMaskedQuoteId( array $payload )
+    public function replaceQuoteIdwithMaskedQuoteId(array $payload)
     {
         $decoded_payload = json_decode($payload, true);
-        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(( $decoded_payload['QuoteId'] ));
+        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(($decoded_payload['QuoteId']));
         unset($decoded_payload['QuoteId']);
 
         return $payload = json_encode(array_merge(
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -7,6 +7,8 @@ use Klaviyo\Reclaim\Model\SyncsFactory;
 use Klaviyo\Reclaim\Model\Quote\QuoteIdMask;
 use Klaviyo\Reclaim\Model\Resourcemodel\Events\CollectionFactory;
 
+use Magento\Catalog\Model\CategoryFactory;
+
 class EventsTopic
 {
     /**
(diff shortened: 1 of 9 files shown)


Cron/EventsTopic.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Cron method used by klaviyo_events_topic cron job
+     * Reads rows from kl_events table and writes to the kl_sync table
+     */
+    public function moveRowsToSync()
+    {
+        // New Events to be moved to kl_sync table and update status of these to Moved, limit 500
+        $eventsCollection = $this->_eventsCollectionFactory->create();
+        $eventsData = $eventsCollection->getRowsForSync('NEW')
+            ->addFieldToSelect(['id','event','payload','user_properties'])
```
ihor-sviziev, line 64: I think there some limits should be applied to process N events for a single run and not to get out of memory

Cron/EventsTopic.php
```diff
@@ -0,0 +1,125 @@
...
+     * This needs to be done here since MaskedQuoteId is unavailable when rows are recorded in kl_events table
+     * @param $payload
+     * @return false|string
+     */
+    public function addMaskedQuoteIdToEventPayload( $payload )
+    {
+        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(( $payload['QuoteId'] ));
+        unset($payload['QuoteId']);
+
+        return $payload = json_encode(array_merge(
```
ihor-sviziev, line 119: From the method name it's not clear that it's going to json encode the result. I think it's better to return array here and to Jain encode in the "client code"

Cron/EventsTopic.php
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Helper method to get MaskedQuoteId from quote_id_mask table and replace QuoteId in payload
+     * This needs to be done here since MaskedQuoteId is unavailable when rows are recorded in kl_events table
+     * @param $payload
+     * @return false|string
+     */
+    public function addMaskedQuoteIdToEventPayload( $payload )
+    {
+        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId(( $payload['QuoteId'] ));
+        unset($payload['QuoteId']);
```
ihor-sviziev, line 117: Looks like method "addMasquedQuoteId" shouldn't remove any data

Cron/EventsTopic.php
```diff
@@ -0,0 +1,125 @@
...
+
+        return $payload = json_encode(array_merge(
+            $payload,
+            array('MaskedQuoteId' => $maskedQuoteId)
```
ihor-sviziev, line 121: Everyone in php now using short syntax for arrays ;)

Cron/KlSyncs.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Sync Cron job sending webhooks and events to Klaviyo
+     * @throws Exception
+     */
+    public function sync()
+    {
+        $syncCollection = $this->_syncCollectionFactory->create();
+        $groupedRows = $this->getGroupedRows( $syncCollection->getRowsForSync('NEW')->getData() );
```
ihor-sviziev, line 64: I guess some limitations should be added there

Cron/KlSyncs.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * Retry Cron job retries all rows marked for retry in kl_sync table
+     * @throws Exception
+     */
+    public function retry()
+    {
+        $syncCollection = $this->_syncCollectionFactory->create();
+        $groupedRows = $this->getGroupedRows( $syncCollection->getRowsForSync('RETRY')->getData() );
```
ihor-sviziev, line 96: And there should be some limitations

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,161 @@
...
+            'AddedItemUrl' => (string) $addedProduct->getProductUrl()
+        );
+
+        $klAddedToCartPayload = array_merge(
+            $this->klBuildCartData( $quote, $addedItem ),
+            $addedItemData
+        );
+
+        // Creating a custom session variable in the checkout session since quoteId is not set at this point for guest checkouts
+        $this->getCheckoutSession()->setKlAddedToCartKey( $klAddedToCartPayload );
```
ihor-sviziev, line 75: I think It's not the best option since after magento 2.4.3 and 2.3.7-p1 we have limit for the session size. Is there any other options?

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Helper function to add items from cart to Added to Cart payload
+     * @param $quote
+     * @param $addedItem
+     * @return array
+     */
+    public function klBuildCartData($quote, $addedItem)
+    {
+        $cartItems = $quote->getItems() ?? array();
```
ihor-sviziev, line 86: Note that configurable products for example are having 2 quote items, one of it is "simple" product that you selected and one of it "parent". I guess you should select the visible items only and (maybe) get some data from the parent item

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,161 @@
...
+     * Retrieves category names from category IDs
+     * @param array $categoryIds
+     * @return array
+     */
+    public function getCategoryName(array $categoryIds)
+    {
+        $categoryFactory = $this->_categoryFactory->create();
+        $categoryNames = [];
+        foreach ( $categoryIds as $id ) {
+            $category = $categoryFactory->load($id);
```
ihor-sviziev, line 132: Load in the loop isn't good idea at all. It will significantly reduce performance of adding to cart in case if product assigned to let's say 100 categories

Cron/EventsTopic.php
siddwarkhedkar: I'll rename the method, we don't want to send the QuoteId over the public API, it is meant to replace it

Cron/KlSyncs.php
```diff
@@ -0,0 +1,189 @@
...
+    public function clean()
+    {
+        $syncCollection = $this->_syncCollectionFactory->create();
+        $idsToDelete = $syncCollection->getIdsToDelete('SYNCED');
+
+        $syncCollection->deleteRows($idsToDelete);
+
+        $klSyncTableSize = $syncCollection->count();
+        if ($klSyncTableSize > 10000 ){
+            $this->_klaviyoLogger->log("Klaviyo Clean Sync: kl_sync table size greater than 10000, currently sitting at $klSyncTableSize");
```
klaviyojad, line 83:
```suggestion
            $this->_klaviyoLogger->log("Warning: kl_sync table size greater than 10000, currently sitting at $klSyncTableSize");
```

Cron/KlSyncs.php
```diff
@@ -0,0 +1,189 @@
...
+        }
+        $this->updateRowStatuses( $responseManifest, $isRetry );
+    }
+
+    /**
+     * Update statues of rows to SYNCED, RETRY and FAILED based on response and if Retry cron run
+     * @param $responseManifest
+     * @param $isRetry
+     */
+    private function updateRowStatuses( $responseManifest, $isRetry )
```
klaviyojad, line 153: whats the idea behind the statuses.  If it fails the first time you assign it 'RETRY' and if second time 'FAILED'?

Helper/Data.php
```diff
@@ -153,9 +155,19 @@ protected function build_params($params) {
         return 'data=' . urlencode(base64_encode(json_encode($params)));
     }
 
+    protected function unwrap_params($params) {
```
klaviyojad, line 158: could you describe what you are doing here?

siddwarkhedkar: Just trying to get it in a readable format to log into the Klaviyo.log file if any failures occur on the track call

Cron/KlSyncs.php
siddwarkhedkar: that's correct

Observer/SalesQuoteProductAddAfter.php
ihor-sviziev: Hi @siddwarkhedkar,
The "add to cart" and "save quote after" are done in a single web request if I got it right.
In this case, it's better to introduce some object that will store this information and won't save data to session and DB till you need it.

Example of such code:
```php
class EventsQueue {
    private $items;

    public function addItem(array $payload): void
    {
        $this->items[] = $payload;
    }

    public function getItems(): array
    {
        return $this->items;
    }
}
```

And client code will be like this:

```php
// in the SalesQuoteProductAddAfter.php
$this->eventsQueue->addItem($klAddedToCartPayload);
```

and
```php
// in the SalesQuoteSaveAfter.php
$this->eventsQueue->getItems();
```

Cron/KlSyncs.php
```diff
@@ -0,0 +1,187 @@
...
+                foreach($rows as $row) {
+                    $response = $this->_dataHelper->klaviyoTrackEvent(
+                        $row['topic'],
+                        json_decode($row['user_properties'], true ),
+                        json_decode($row['payload'], true ),
+                        strtotime($row['created_at'])
+                    );
+                    if (!$response) {$response = '0';}
+
+                    array_push($responseManifest["$response"], $row['id']);
```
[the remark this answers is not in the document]
siddwarkhedkar: The condition in which the klaviyoTrackEvent method will return that string is taken care of in the SalesQuoteSaveAfter where we check if the customer is identified, it is fairly unlikely for that string to be returned since a row in the DB will not exist without the customer properties

Helper/Data.php
```diff
@@ -153,9 +161,19 @@ protected function build_params($params) {
...
+        return json_decode(base64_decode(urldecode(substr($params,5))), true);
+    }
+
     protected function make_request($path, $params) {
         $url = self::KLAVIYO_HOST . $path . '?' . $params;
         $response = file_get_contents($url);
+
+        if ($response == '0'){
+            $dataString = print_r($this->unwrap_params($params),true);
+            $this->_klaviyoLogger->log("Unable to send event to Track API with data: $dataString");
```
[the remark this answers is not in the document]
siddwarkhedkar: Hoping it's going to be a rare instance that a long string is logged to the klaviyo.log, but taking a look at my testing instance, all other files in log directory tend to get large and doesn't affect the store. 
The point of logging is that Product Experts are easily able to retrieve the payload and check what may not be as expected. 
Discussed offline too, resolving

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,189 @@
...
+        foreach ($addedItems as $item){
+            if ($item->getProductType() == 'bundle') {
+                $children = $item->getChildren();
+                foreach ($children as $child){
+                    array_push($childrenIds, $child->getId());
+                }
+            }
+
+            if (in_array( $item->getId(), $childrenIds )){
+                return;
```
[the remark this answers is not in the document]
siddwarkhedkar: My bad, supposed to be continue

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,189 @@
...
+
+        foreach ($addedItems as $item){
+            if ($item->getProductType() == 'bundle') {
+                $children = $item->getChildren();
+                foreach ($children as $child){
+                    array_push($childrenIds, $child->getId());
+                }
+            }
+
+            if (in_array( $item->getId(), $childrenIds )){
```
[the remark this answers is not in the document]
siddwarkhedkar: Going to add a condition to account for this. However, when a bundled item is added, there parent item is the first in the array followed by children, what I've seen in testing. 

But still going to account for this condition to avoid surprises on deployment

Observer/SalesQuoteProductAddAfter.php
siddwarkhedkar: @ihor-sviziev I'm moving the hydration categoryNames to the cron so as to ensure heavy lifting is done later without affecting end user experience

Observer/SalesQuoteSaveAfter.php
```diff
@@ -0,0 +1,87 @@
...
+    public function execute(Observer $observer)
+    {
+        // Checking if the cookie is set here, if not it will return undefined and break the code
+        if ( !isset($_COOKIE['__kla_id'] )) { return; }
+        $kl_decoded_cookie = json_decode(base64_decode($_COOKIE['__kla_id']), true);
+
+        // Get the custom variable set in the DataHelper object via the SalesQuoteProductAddAfter observer.
+        // Check if the public key and Added to Cart payload are set
+        $public_key = $this->_scopeSetting->getPublicApiKey();
+        $klAddedToCartPayload = $this->_dataHelper->tempPayload;
```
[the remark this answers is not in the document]
siddwarkhedkar: https://github.com/klaviyo/magento2-klaviyo/pull/142#discussion_r730735125

Cron/EventsTopic.php
```diff
@@ -0,0 +1,190 @@
...
+     * @return array
+     */
+    public function searchCategoryMapAndReturnNames(array $categoryIds, array $categoryMap): array
+    {
+        $categoryNames = [];
+        foreach ($categoryIds as $id){
+            array_push($categoryNames, $categoryMap[$id]);
+        }
+
+        return $categoryNames;
```
klaviyojad, line 188:
```suggestion    
        return array_filter($categoryIds, function($id) {
            return $categoryMap[$id];
        });
```

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,183 @@
...
+
+        foreach ($addedItems as $item){
+            if ($item->getProductType() == 'bundle') {
+                $children = $item->getChildren();
+                foreach ($children as $child){
+                    if (!in_array($child->getId(), $childrenIds)) {
+                        array_push($childrenIds, $child->getId());
+                    }
+                }
+            } elseif (!is_null($item->getParentItem()) && !in_array($item->getId(), $childrenIds)) { //If child comes  before Parent, collect those too
```
klaviyojad, line 55: why is this even performed.  can we just do nothing if the above if condition is the case after all elseifs are satisfied?

Cron/EventsTopic.php
```diff
@@ -0,0 +1,190 @@
...
+
+        if (empty($eventsData)){
+            return;
+        }
+
+        $idsMoved = [];
+
+        // Capture all events that have been moved and add data to Sync table
+        foreach ($eventsData as $event){
+            if ($event['topic'] == 'Added To Cart') {
```
klaviyojad, line 85: I think this is a bug, kl_events table doesn't have column named 'topic'
```suggestion
            if ($event['event'] == 'Added To Cart') {
```

Cron/EventsTopic.php
ihor-sviziev: It looks like not an equal to what was there before. I guess you wanted to write here array_map? :)

klaviyojad: yea my code above fails, I guess I was just suggesting if you can find a lambda way of getting the end result

Cron/KlSyncs.php
```diff
@@ -0,0 +1,186 @@
...
+        {
+            if (array_key_exists($row['topic'], $groupedRows))
+            {
+                array_push($groupedRows[$row['topic']], $row);
+            }
+            else
+            {
+                $groupedRows[$row['topic']] = [$row];
+            }
+        }
```
smoucka, line 182:
```suggestion
        foreach ($allRows as $row) {
            if (array_key_exists($row['topic'], $groupedRows)) {
                array_push($groupedRows[$row['topic']], $row);
            } else {
                $groupedRows[$row['topic']] = [$row];
            }
        }
```

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -0,0 +1,183 @@
...
+    public function execute(Observer $observer)
+    {
+        $quote = $observer->getData('items')[0]->getQuote();
+        $addedItems = $observer->getData('items');
+
+        // Create a list of Simple Product Ids being added as part of a bundle.
+        // We collect the Item name and Qty added to cart, to be sent as AddedItemBundleOptions with the payload.
+        $childrenIds = [];
+
+        foreach ($addedItems as $item){
```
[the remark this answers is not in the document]
siddwarkhedkar: With bundled products and grouped products, more than 1 item is added to the cart and sent over with the observer. 

For grouped products > all are added separately to the cart but come as 3 different products into the observer, so we need to iterate over each create a payload for each

for bundled products > parent item and children are sent with the observer but only the parent item is shown in the cart, so we don't want to process a payload for the children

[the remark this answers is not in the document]
smoucka: Cool, so in the case of a grouped product with 3 simple products only a single observer event fires with all of the products. We won't get 3 observer events each with 3 simple products?

Cron/EventsTopic.php
```diff
@@ -0,0 +1,190 @@
...
+                'topic' => $event['event'],
+                'user_properties' => $event['user_properties'],
+                'payload' => $event['payload']
+            ]);
+            try {
+                $sync->save();
+                array_push($idsMoved, $event['id']);
+            } catch (\Exception $e) {
+                $this->_klaviyoLogger->log(sprintf("Unable to move row: %s", $e));
+            }
```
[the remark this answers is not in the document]
siddwarkhedkar: Didn't get a way to trigger it since it's to catch any exception thrown during row transfer. 
But yes, getMessage() would be the better

Cron/EventsTopic.php
```diff
@@ -0,0 +1,190 @@
...
+    public function getCategoryMap(array $categoryIds): array
+    {
+        $categoryFactory = $this->_categoryFactory->create();
+        $categoryMap = [];
+
+        foreach($categoryIds as $categoryId){
+            $categoryMap[$categoryId] = $categoryFactory->load($categoryId)->getName();
+        }
+
+        return $categoryMap;
```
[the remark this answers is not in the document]
siddwarkhedkar: I could create a class property to store the map, such that we only load a category once and only when it's name is not stored in the map

Observer/SalesQuoteProductAddAfter.php
siddwarkhedkar: I guess we could return, but further down in the loop iteration we basically build the cart payload and we don't want to do that for any children Ids, that is why I'm adding it to the array if it has a parentItem associated

Cron/EventsTopic.php
siddwarkhedkar: Found a better way to achieve this, coming up in next commit

smoucka (approved):

Cron/EventsTopic.php
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Helper function to replace QuoteId with MaskedQuoteId and CategoryIds with CategoryNames in Added To Cart payload
+     * @param string $payload
+     * @return array
+     */
+    public function replaceQuoteIdAndCategoryIds(string $payload): array
+    {
+        $decoded_payload = json_decode($payload, true);
+        $maskedQuoteId = $this->_quoteIdMaskResource->getMaskedQuoteId($decoded_payload['QuoteId']);
+        $decoded_payload['MaskedQuoteId'] = $maskedQuoteId;
```
ihor-sviziev, line 136: the masked quote id is available only for guest carts AFAIK, so probably it will just remove the quote id for registered customers

PR #144, opened 2021-11-08
Adds product/save to message queue

jordanallain:
- Adds a getWebhooks method/endpoint to the Reclaim interface for
querying the customizable webhooks.
- Adds the observer that runs when a product is saved.
- Adds the Cron class for managing product data that needs to be synced
to Klaviyo.
- Admin HTML for configuring the webhook.
- Modify the webhook helper so we only encode the json once.

### DIFF
--- a/Cron/ProductsTopic.php
+++ b/Cron/ProductsTopic.php
@@ -1,0 +1,96 @@
+<?php
+
+namespace Klaviyo\Reclaim\Cron;
+
+use Klaviyo\Reclaim\Helper\Logger;
+use Klaviyo\Reclaim\Model\SyncsFactory;
+use Klaviyo\Reclaim\Model\ResourceModel\Products;
+use Klaviyo\Reclaim\Model\ResourceModel\Products\CollectionFactory;
+
+class ProductsTopic
+{
+    /**
+     * Klaviyo Logger
+     * @var Logger
+     */
+    protected $_klaviyoLogger;
+
+    /**
+     * Klaviyo Products Resource Model
+     * @var Products
+     */
+    protected $_klProduct;
+
+    /**
+     * Klaviyo Products Collection
+     * @var CollectionFactory
+     */
+    protected $_klProductCollectionFactory;
+
+    /**
+     * Klaviyo Syncs Model
+     * @var SyncsFactory
+     */
+    protected $_klSyncFactory;
+
+    /**
+     * @param Logger $klaviyoLogger
+     * @param Products $klProduct
+     * @param SyncsFactory $klSyncFactory
+     * @param CollectionFactory $klProductCollectionFactory
+     */
+    public function __construct(
+        Logger $klaviyoLogger,
+        Products $klProduct,
+        SyncsFactory $klSyncFactory,
+        CollectionFactory $klProductCollectionFactory
+    )
+    {
+        $this->_klaviyoLogger = $klaviyoLogger;
+        $this->_klProduct = $klProduct;
+        $this->_klSyncFactory = $klSyncFactory;
+        $this->_klProductCollectionFactory = $klProductCollectionFactory;
+    }
+
+    public function queueKlProductsForSync()
+    {
+        $klProductsCollection = $this->_klProductCollectionFactory->create();
+        $klProductsToSync = $klProductsCollection->getRowsForSync('NEW')
+            ->addFieldToSelect(['id','payload','status','topic', 'klaviyo_id'])
+            ->getData();
+
+        if (empty($klProductsToSync))
+        {
+            return;
+        }
+
+        $idsToUpdate = [];
+
+        foreach ($klProductsToSync as $klProductToSync)
+        {
+            $klSync = $this->_klSyncFactory->create();
+            $klSync->setData([
+                'payload'=> $klProductToSync['payload'],
+                'topic'=> $klProductToSync['topic'],
+                'klaviyo_id'=>$klProductToSync['klaviyo_id'],
+                'status'=> 'NEW'
+            ]);
+            try {
+                $klSync->save();
+                array_push($idsToUpdate, $klProductToSync['id']);
+            } catch (\Exception $e) {
+                $this->_klaviyoLogger->log(sprintf('Unable to move row: %s', $e));
+            }
+        }
+
+        $klProductsCollection->updateRowStatus($idsToUpdate, 'MOVED');
+    }
+
+    public function clean()
+    {
+        $klProductsCollection = $this->_klProductCollectionFactory->create();
+        $idsToDelete = $klProductsCollection->getIdsToDelete('MOVED');
+
+        $klProductsCollection->deleteRows($idsToDelete);
+    }
+}
--- a/Helper/ScopeSetting.php
+++ b/Helper/ScopeSetting.php
@@ -30,7 +30,8 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
 
     const WEBHOOK_SECRET = 'klaviyo_reclaim_webhook/klaviyo_webhooks/webhook_secret';
     const PRODUCT_DELETE_BEFORE = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_delete_before_webhook';
-    
+    const PRODUCT_SAVE_AFTER = 'klaviyo_reclaim_webhook/klaviyo_webhooks/using_product_save_after_webhook';
+
     const KLAVIYO_OAUTH_NAME = 'klaviyo_reclaim_oauth/klaviyo_oauth/integration_name';
 
     protected $_scopeConfig;
@@ -143,6 +144,19 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
         return $this->getScopeSetting(self::WEBHOOK_SECRET, $storeId);
     }
 
+    public function getWebhooks()
+    {
+        $registeredWebhooks = [];
+
+        $product_delete = $this->getProductDeleteBeforeSetting();
+        $product_save = $this->getProductSaveAfterSetting();
+
+        array_push($registeredWebhooks, ['product/delete', $product_delete)];
+        array_push($registeredWebhooks, ['product/save', $product_save]);
+
+        return $registeredWebhooks;
+    }
+
     public function isEnabled($storeId = null)
     {
         return $this->getScopeSetting(self::ENABLE, $storeId);
@@ -221,7 +235,7 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
     {
         return $this->getScopeSetting(self::CONSENT_AT_CHECKOUT_SMS_LIST_ID, $storeId);
     }
-    
+
     public function getConsentAtCheckoutSMSConsentText($storeId = null)
     {
         return $this->getScopeSetting(self::CONSENT_AT_CHECKOUT_SMS_CONSENT_TEXT, $storeId);
@@ -264,5 +278,9 @@ class ScopeSetting extends \Magento\Framework\App\Helper\AbstractHelper
         return $this->getScopeSetting(self::PRODUCT_DELETE_BEFORE, $storeId);
     }
 
-}
+    public function getProductSaveAfterSetting($storeId = null)
+    {
+        return $this->getScopeSetting(self::PRODUCT_SAVE_AFTER, $storeId);
+    }
 
+}
--- a/Helper/Webhook.php
+++ b/Helper/Webhook.php
@@ -33,14 +33,13 @@ class Webhook extends \Magento\Framework\App\Helper\AbstractHelper
 
     /**
      * @param string $webhookType
-     * @param array $data
+     * @param string $data
      * @param string $klaviyoId
      * @return string
      * @throws Exception
      */
     public function makeWebhookRequest($webhookType, $data, $klaviyoId=null)
     {
-
         if (!$klaviyoId) {
             $klaviyoId = $this->_klaviyoScopeSetting->getPublicApiKey();
         }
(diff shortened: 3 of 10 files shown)


Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,150 @@
...
+                "status"=>"NEW",
+                "topic"=>"product/save",
+                "klaviyo_id"=>$klaviyoId,
+                "payload"=>json_encode($normalizedProduct)
+              ];
+              $klProduct = $this->_klProductFactory->create();
+              $klProduct->setData($data);
+              $klProduct->save();
+
+              // $this->_klaviyoLogger->log( print_r("Created new row?", true));
```
jordanallain, line 96: remove 😑

jordanallain, line 96: remove 😑

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,150 @@
...
+
+              // $this->_webhookHelper->makeWebhookRequest('product/save', $normalizedProduct, $klaviyoId);
```
jordanallain, line 98: remove 😑

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,150 @@
...
+
+    private function normalizeProduct($product=null)
+    {
+      if ($product == null) {
+        return;
+      }
+
+      $product_id = $product->getId();
+
+      // remove thumnbail image url?
```
jordanallain, line 111: remove 😑

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,139 @@
...
+    protected $_klaviyoLogger;
+
+    /**
+     * @var Webhook $webhookHelper
+     */
+    protected $_webhookHelper;
+    protected $_categoryFactory;
+    protected $_klProductFactory;
+    protected $_stockRegistry;
+    protected $product_category_names = [];
```
siddwarkhedkar, line 39: We could remove a few unused protected variables and add variable descriptions to each

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,139 @@
...
+    public function execute(Observer $observer)
+    {
+        $product = $observer->getEvent()->getProduct();
+        $storeIds = $product->getStoreIds();
+        $storeIdKlaviyoMap = $this->_klaviyoScopeSetting->getStoreIdKlaviyoAccountSetMap($storeIds);
+
+        foreach ($storeIdKlaviyoMap as $klaviyoId => $storeIds) {
+            if (empty($storeIds)) {continue;}
+
+            if ($this->_klaviyoScopeSetting->getWebhookSecret() && $this->_klaviyoScopeSetting->getProductSaveAfterSetting($storeIds[0])) {
```
siddwarkhedkar, line 81: to keep consistency with previous suggestion
```suggestion
            if ($this->_klaviyoScopeSetting->getWebhookSecret() && $this->_klaviyoScopeSetting->getProductSaveWebhookSetting($storeIds[0])) {
```

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,139 @@
...
+              $normalizedProduct = $this->normalizeProduct($product);
+              $data = [
+                "status"=>"NEW",
+                "topic"=>"product/save",
+                "klaviyo_id"=>$klaviyoId,
+                "payload"=>json_encode($normalizedProduct)
+              ];
+              $klProduct = $this->_klProductFactory->create();
+              $klProduct->setData($data);
+              $klProduct->save();
```
siddwarkhedkar, line 91: I don't think we need to create a new Model to save a new row
```suggestion
              $klProduct = $this->_klProductFactory->setData($data);
              $klProduct->save();
```

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,139 @@
...
+        $product_info['metadata']['special_from_date'] = $product->getSpecialFromDate();
+        $product_info['metadata']['special_to_date'] = $product->getSpecialToDate();
+      }
+
+      $product_category_ids = $product->getCategoryIds();
+      $category_factory = $this->_categoryFactory->create();
+      foreach ($product_category_ids as $category_id) {
+        $category = $category_factory->load($category_id);
+        $product_info['product']['categories'][$category_id] = $category->getName();
+      }
```
siddwarkhedkar, line 135: Following up on this comment, https://github.com/klaviyo/magento2-klaviyo/pull/142#discussion_r724678177
Maybe it would be wise to move the categoryName population to the Cron job to safeguard from potential efficiency issues for products which fall in a ton of categories

etc/adminhtml/system.xml
```diff
@@ -158,6 +158,11 @@
                     <source_model>Magento\Config\Model\Config\Source\Yesno</source_model>
                     <comment>This will remove deleted products from the Klaviyo catalog.</comment>
                 </field>
+                <field id="using_product_save_after_webhook" translate="label" type="select" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
```
siddwarkhedkar, line 161:
```suggestion
                <field id="using_product_save_webhook" translate="label" type="select" sortOrder="30" showInDefault="1" showInWebsite="1" showInStore="1">
```

siddwarkhedkar (commented):
Excited to have the last piece follow through! Looks great, I had a few general suggestions which I pulled from the reviews on the previous PRs to maintain consistency

Helper/ScopeSetting.php
```diff
@@ -143,6 +144,19 @@ public function getWebhookSecret($storeId = null)
...
 
+    public function getWebhooks()
+    {
+        $registeredWebhooks = [];
+
+        $product_delete = $this->getProductDeleteBeforeSetting();
+        $product_save = $this->getProductSaveAfterSetting();
+
+        array_push($registeredWebhooks, ['product/delete', $product_delete]);
+        array_push($registeredWebhooks, ['product/save', $product_save]);
```
klaviyojad, line 155: I am seeing [somewhere](https://appdividend.com/2019/03/29/php-array-push-example-add-elements-to-an-array/) that this is possible:
```suggestion
        array_push($registeredWebhooks, ['product/delete', $product_delete]), ['product/save', $product_save]);
```

Helper/Webhook.php
```diff
@@ -33,14 +33,13 @@ public function __construct(
 
     /**
      * @param string $webhookType
-     * @param array $data
+     * @param string $data
```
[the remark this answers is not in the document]
siddwarkhedkar: Maybe we could add a line to ensure that this method expects a json_encoded string?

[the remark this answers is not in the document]
jordanallain: do you mean as a comment or as actual code?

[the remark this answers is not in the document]
siddwarkhedkar: I mean as a comment in the methods docstring

Observer/ProductSaveAfter.php
jordanallain: i tried this suggested change but i got an error when it ran.

siddwarkhedkar: scratch that, you're correct, my bad on this one!

Cron/ProductsTopic.php
```diff
@@ -0,0 +1,123 @@
...
+
+        $klProductsCollection->deleteRows($idsToDelete);
+    }
+
+    /**
+     * Helper function to associate category names with their ids
+     * @param string $payload
+     * @return array
+     */
+    public function addCategoryNames(string $payload): array
```
siddwarkhedkar, line 110: Over here, we will be iterating over each categoryId for all payloads passed, so we will be running the for loop 500 times, if categoryIds, exist and possibly loading duplicate categroyModels since it is likely more than one product will have a category repeated. Can we make it such that we create a class level attribute to have Ids and categoryNames associated so don't duplicate the effort? 

https://github.com/klaviyo/magento2-klaviyo/pull/142#discussion_r742024644 > as advised here? 

Or do you think it is worth having a method in the helper class which can be used by both the Cron?

Helper/Webhook.php
```diff
@@ -66,24 +65,23 @@ public function makeWebhookRequest($webhookType, $data, $klaviyoId=null)
         $err = curl_errno($curl);
 
         if ($err) {
-            $this->_klaviyoLogger->log(sprintf('Unable to send webhook to %s with data: %s', $url, json_encode($data)));
+            $this->_klaviyoLogger->log(sprintf("Unable to send webhook to $url with data: $data"));
```
siddwarkhedkar, line 68: super nit:
```suggestion
            $this->_klaviyoLogger->log("Unable to send webhook to $url with data: $data");
```

Helper/Webhook.php
```diff
@@ -66,24 +65,23 @@ public function makeWebhookRequest($webhookType, $data, $klaviyoId=null)
...
 
         // Close cURL session handle
         curl_close($curl);
+
         return $response;
     }
 
     /**
-     * @param array data
+     * @param string data
```
siddwarkhedkar, line 78: Could we add a description to this method?

Observer/ProductSaveAfter.php
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    private function normalizeProduct($product=null)
+    {
+      if ($product == null) {return;}
+
+      $product_id = $product->getId();
+
+      $product_info = array(
+        'store_ids' => $product->getStoreIds(),
```
siddwarkhedkar, line 88: Are all array_keys supposed to start lower case/upper case? Could we standardize across the payload?

siddwarkhedkar (commented):
Some thoughts about how loading categories and a couple of nits, looks great otherwise!

Observer/ProductSaveAfter.php
jordanallain: I only use uppercase letters on some of the properties because Klaviyo wants them that way so this avoids normalizing the payload in the app. otherwise I user lowercase where I can. Let me know if that's alright or not!

Helper/Webhook.php
jordanallain: Does this work?

```
* @param string $data json payload used to create hmac signature
```

Helper/Webhook.php
jordanallain: whoops haha!

Helper/Webhook.php
siddwarkhedkar:
```suggestion
     * Returns an HMAC signature for webhooks
     * @param string data
```

Observer/ProductSaveAfter.php
siddwarkhedkar: Is there any restriction on making other start upper case too? Just looking for standardize keys. it's a nitpick so if it is too much work to accommodate this, we can skip it

Helper/ScopeSetting.php
```diff
@@ -143,6 +144,14 @@ public function getWebhookSecret($storeId = null)
         return $this->getScopeSetting(self::WEBHOOK_SECRET, $storeId);
     }
 
+    public function getWebhooks()
+    {
+        return $registeredWebhooks = [
+          ['product/delete', $this->getProductDeleteWebhookSetting()],
+          ['product/save', $this->getProductSaveWebhookSetting()],
+        ];
```
smoucka, line 152: Is there value in returning topics that aren't enabled? Thinking about other platforms for which we subscribe to certain webhooks (Shopify, WooCommerce), I'm pretty sure the endpoints that return this info only return the topics to which we are registered rather than a comprehensive list of all available.

I think it might be cleaner (and more easily extended in the future) to format this as an indexed array of associative arrays, similar to what those other platforms return. ex:
```
[
    [
        'topic' => 'product/delete',
        'enabled' => $this->getProductDeleteWebhookSetting(),
    ],
    [
        'topic' => 'product/save',
        'enabled' => $this-> getProductSaveWebhookSetting(),
    ]
]

Comment (smoucka) [reply Helper/ScopeSetting.php:152]: We could also allow for a param in this request to optionally fetch all available webhooks vs those that are enabled

Comment (jordanallain) [reply Helper/ScopeSetting.php:152]: I'd be fine with only sending back only the enabled webhooks if we want to go that direction. In the app code that hits this endpoint I essentially filter out the ones that aren't enabled anyway. If we're only going to return the enabled ones then I'm not sure we need anything other than a list of the topics though?

Comment (smoucka) [reply Helper/ScopeSetting.php:152]: Yea, I guess since this is only for an inspector task it's probably fine however. Is there any other info stored about this? like when this setting was last updated for example? That could be useful for troubleshooting but may not be available

Comment (smoucka) [reply Helper/ScopeSetting.php:152]: Ahh, forgot about that. In that case if it's used in two different scenarios, I'd let the app code contain the logic to process the response/format the value to be saved and return as much info here for troubleshooting purposes which can be accessed via the inspector task.

Comment (smoucka) [reply Helper/ScopeSetting.php:152]: I'd love to see this return an indexed array containing an associative array for each webhook topic. Makes the logic in the app way easier to read instead of relying on the index for the topic and enabled boolean.

Comment (jordanallain) [reply Helper/ScopeSetting.php:152]: did you see the updated logic for that method in the app? it still relies on the index but it unpacks the values into named variables which i think helps improve the readability.

```py
for webhook in webhook_config:
    webhook_name, is_enabled = webhook
    if int(is_enabled):
        registered_webhooks.append(webhook_name)
```

if i switch this back to an associative array, did you have logic in mind to pull out only the enabled webhooks' name into a list that would be more readable?

jordanallain: sorry for the confusion around this! should be updated now though to what you were describing.

siddwarkhedkar (approved):
:shipit: LGTM!

PR #154, opened 2022-01-04
Update cart.phtml

JamesFX2:
Fixes for us, somehow being called before this is set in other places.

### DIFF
--- a/view/frontend/templates/checkout/cart.phtml
+++ b/view/frontend/templates/checkout/cart.phtml
@@ -4,6 +4,7 @@
            'jquery',
            'mage/url'
         ], function (jQuery, url) {
+          url.setBaseUrl(window.BASE_URL);
           jQuery.ajax({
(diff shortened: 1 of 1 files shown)


siddwarkhedkar:
Hey @JamesFX2 , What is the issue you are facing and how does this fix work? Could you please provide a summary of why this is necessary? 

It will be great if you could also mention what version of the extension you are using, what version of M2 you are on

JamesFX2:
Latest version of your extension. V2.3.5

Not sure if this is being caused by MagePack or your extension but it's calling /checkout/cart/reclaim/checkout/reload (404) on our release site's cart page. I guess it's being called before the baseUrl is set. 

I've added an override in our theme that does this so your call if you take it or not.

siddwarkhedkar:
Hey @JamesFX2 , I'm going to close out this PR and add this to our backlog to take a look into this soon. After testing, we'll decide if we want to make the change to any future extensions

PR #155, opened 2022-01-05
Rebuild cart for Logged-in users, whitelist klaviyo scripts, identify logged in users correctly

siddwarkhedkar:
This PR fixes a few things: 

MaskedQuoteId is not generated for quotes associated with Logged in users. In such a situation, we would send a null back with the ATC payload, with no way to rebuild cart for these users. 
[TP](https://klaviyo.tpondemand.com/entity/99989-rebuild-carts-for-logged-in-customers)

There was a lot of noise in the console caused by csp policies not whitelising the klaviyo scripts
[TP](https://klaviyo.tpondemand.com/entity/101804-whiteslist-klaviyo-domains-in-magento-2)

Logged in users were not correctly being cookied as the customerData was being generated after script. It will now wait for this to be generated before trying to cookie the user.

### DIFF
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -46,7 +46,7 @@ class Cart extends \Magento\Framework\App\Action\Action
 
         unset($params['quote_id']);
 
-        // Check if the quote_id has kx_identifier, if yes, retreive active quote for customer, if not get QuoteId from masked QuoteId
+        // Check if the quote_id has kx_identifier, if yes, retrieve active quote for customer, if not get QuoteId from masked QuoteId
         if (strpos($quoteId, "kx_identifier_") !== false){
             $customerId = base64_decode( str_replace("kx_identifier_", "", $quoteId) );
             try {
--- a/Controller/Checkout/Cart.php
+++ b/Controller/Checkout/Cart.php
@@ -3,6 +3,8 @@
 namespace Klaviyo\Reclaim\Controller\Checkout;
 
 
+use Magento\Framework\Exception\NoSuchEntityException;
+
 class Cart extends \Magento\Framework\App\Action\Action
 {
     protected $quoteRepository;
@@ -44,12 +46,24 @@ class Cart extends \Magento\Framework\App\Action\Action
 
         unset($params['quote_id']);
 
-        try {
-            $quoteIdMask = $this->quoteIdMaskFactory->create()->load($quoteId, 'masked_id');
-            $quote = $this->quoteRepository->get($quoteIdMask->getQuoteId());
-            $this->cart->setQuote($quote);
-            $this->cart->save();
-        } catch (\Magento\Framework\Exception\NoSuchEntityException $ex) {
+        // Check if the quote_id has kx_identifier, if yes, retreive active quote for customer, if not get QuoteId from masked QuoteId
+        if (strpos($quoteId, "kx_identifier_") !== false){
+            $customerId = base64_decode( str_replace("kx_identifier_", "", $quoteId) );
+            try {
+                $quote = $this->quoteRepository->getActiveForCustomer($customerId);
+                $this->cart->setQuote($quote);
+                $this->cart->save();
+            } catch (NoSuchEntityException $ex){
+
+            }
+        } else {
+            try {
+                $quoteIdMask = $this->quoteIdMaskFactory->create()->load($quoteId, 'masked_id');
+                $quote = $this->quoteRepository->get($quoteIdMask->getQuoteId());
+                $this->cart->setQuote($quote);
+                $this->cart->save();
+            } catch (NoSuchEntityException $ex) {
+            }
         }
 
         $redirect = $this->resultRedirectFactory->create();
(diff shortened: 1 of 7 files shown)


Controller/Checkout/Cart.php
```diff
@@ -44,12 +46,24 @@ public function execute()
 
         unset($params['quote_id']);
 
-        try {
-            $quoteIdMask = $this->quoteIdMaskFactory->create()->load($quoteId, 'masked_id');
-            $quote = $this->quoteRepository->get($quoteIdMask->getQuoteId());
-            $this->cart->setQuote($quote);
-            $this->cart->save();
-        } catch (\Magento\Framework\Exception\NoSuchEntityException $ex) {
+        // Check if the quote_id has kx_identifier, if yes, retreive active quote for customer, if not get QuoteId from masked QuoteId
```
klaviyojad, line 49:
```suggestion
        // Check if the quote_id has kx_identifier, if yes, retrieve active quote for customer, if not get QuoteId from masked QuoteId
```

Observer/SalesQuoteSaveAfter.php
```diff
@@ -68,7 +78,8 @@ public function execute(Observer $observer)
         // Setting QuoteId at this point since the MaskedQuoteId is not updated when this event is dispatched,
         // MaskedQuoteId is set into the payload while the EventsTopic cron job moves rows into the Sync table
         $quote = $observer->getData('quote');
-        $klAddedToCartPayload = array_merge($klAddedToCartPayload, ['QuoteId' => $quote->getId()]);
+        $customerId = $this->checkCustomerAndReturnId($quote);
+        $klAddedToCartPayload['QuoteId'] = isset($customerId) ? "kx_identifier_$customerId" : $quote->getId();
```
klaviyojad, line 82: Im confused, does this make that every added to cart event for a logged in customer with always have the same QuoteId?

siddwarkhedkar: No, just using double quotes to parse the customerId variable and concatenate the string

klaviyojad: Im not asking about the string formatting.  I am seeing that if we have a customer with ID=2 this will always become "kx_identifier_2" whether they added to cart this morning, this afternoonm and even a week from now

Observer/SalesQuoteSaveAfter.php
```diff
@@ -68,7 +78,8 @@ public function execute(Observer $observer)
         // Setting QuoteId at this point since the MaskedQuoteId is not updated when this event is dispatched,
         // MaskedQuoteId is set into the payload while the EventsTopic cron job moves rows into the Sync table
         $quote = $observer->getData('quote');
-        $klAddedToCartPayload = array_merge($klAddedToCartPayload, ['QuoteId' => $quote->getId()]);
+        $customerId = $this->checkCustomerAndReturnId($quote);
```
klaviyojad, line 81: I would rename this encodedCustomerId because that is what it is no?

Observer/SalesQuoteSaveAfter.php
cykolln: Ah I see you can do it that way because it's at the end of the string... learning new things every day about php

Controller/Checkout/Cart.php
```diff
@@ -44,12 +46,24 @@ public function execute()
...
+        if (strpos($quoteId, "kx_identifier_") !== false){
+            $customerId = base64_decode( str_replace("kx_identifier_", "", $quoteId) );
+            try {
+                $quote = $this->quoteRepository->getActiveForCustomer($customerId);
```
klaviyojad, line 53: is it possible for a customer to have more than one active quote at a time? like if they opened 2 different sessions one on their phone and one on their laptop?

Observer/SalesQuoteSaveAfter.php
siddwarkhedkar: @klaviyojad That's right, we'll get the active quote for that specific customer.

I checked this by having multiple incognito windows and starting a session in one place, logging in into the other window and check, Magento populates the contents from the previous quote when the user is logged in so in essence, even if they are on their phone, it would essentially be the same quote if they are logged in to the same account

@cykolln you can use it anywhere, it doesn't need to be at the end of the string. It could also be used as "pronoun$customerId is a customer"

Controller/Checkout/Cart.php
siddwarkhedkar: I answered it in the above comment

klaviyojad (approved):

Observer/SalesQuoteSaveAfter.php
```diff
@@ -96,7 +96,12 @@ public function execute(Observer $observer)
         $this->_dataHelper->unsetObserverAtcPayload();
     }
 
-    private function checkCustomerAndReturnId($quote) {
+    /**
+     * Check if customer isLoggedIn and return base64 encoded string for the ID
```
klaviyojad, line 100:
```suggestion
     * Check if customer is logged in and return base64 encoded string for the ID
```

PR #159, opened 2022-01-19
Get rid of old Install/Upgrade scripts, migrate to declarative schema completely.

siddwarkhedkar:
PR to mitigate Technical review failures: 
image

Removing the Install/Upgrade scripts since we have already moved to Declarative schema. 

For backward compatibility, 1 Patch script for versions older than 1.7 added. 

Created a db_schema_whitelist.json as recommended [here](https://devdocs.magento.com/guides/v2.4/extension-dev-guide/declarative-schema/migration-commands.html?itm_source=devdocs&itm_medium=quick_search&itm_campaign=federated_search&itm_term=whitelis#create-whitelist)


Setup/Patch/Data/UpdateOldPrivateKeysToEncryptedVersions.php
```diff
@@ -0,0 +1,104 @@
...
+     * @inheritdoc
+     */
+    public static function getDependencies()
+    {
+        return [];
+    }
+
+    public static function getVersion()
+    {
+        return '1.1.7';
```
smoucka, line 85: I see the docstring for the method below mentions something about version 1.1.7 which I assume is what this corresponds to. If so is there any other info you can add int eh docstring for how this is utilized? Or does this version number need to be updated?

siddwarkhedkar: The Patch script will only run if the version is less than what is specified in this method. The context below is all that is needed I suppose. If we need to make new changes to any Data, we will require to create a new patch script in the directory corresponding to the version just before the release planned for the patch script

siddwarkhedkar: This is more of a backward compatibility thing required by Magento for migration to Declarative schema

smoucka (approved):
👍 👍

cykolln (approved):

PR #161, opened 2022-01-27
Add store scoping to Added to cart track requests

siddwarkhedkar:


Cron/KlSyncs.php
```diff
@@ -129,11 +129,19 @@ private function sendUpdatesToApp(array $groupedRows, bool $isRetry = false)
                     $eventTime = $decodedPayload['time'];
                     unset($decodedPayload['time']);
 
+                    $storeId = '';
+                    if (isset($decodedPayload['StoreId']))
```
klaviyojad, line 133: Im just curious is there any possible scenario that StoreId wouldn't be set?

siddwarkhedkar: In case this change is made live and some rows have been created before the change was deployed, and a few rows are pending sync to Klaviyo, these won't have the storeId key in the payload

klaviyojad: I dunno if its worth putting a comment or TODO that its only there for backwards compatibility and maybe in the future we just remove it?

siddwarkhedkar: Yup, good thought, I'll add a todo

klaviyojad (approved):

margekar-klaviyo (approved):
Sidd walked me through all the changes and affected components! Changes LGTM!

PR #195, opened 2022-06-28
Fix 404 AJAX Request

jlenzstaysafe:
The base URL needs to be set, otherwise you will generate a relative URL which will 404 on any URL that has a path with more than 1 level deep (ie:  /checkout/cart)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ### [Unreleased]
+ - Fix 404 AJAX Request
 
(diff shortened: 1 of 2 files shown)


cykolln:
This looks good to me and has been tested on our end. Can you please update the CHANGELOG.md and add this fix into the Unreleased section. We typically do releases on a monthly cadence so this will be included in July's release.

jlenzstaysafe:
Sorry, just to clarify, you want me to add a new section like so?  I'm not sure on the date or version number:

### [4.0.5] - 2022-07-24
- Fix 404 AJAX Request

jlenzstaysafe:
Just to clarify, is this the format/text you're looking for? ie:

### [Unreleased]
- Fix 404 AJAX Request

cykolln (approved):

cykolln:
Merged - this should be released in the next two weeks bundled with some other bug fixes. Thanks for your contribution!

PR #198, opened 2022-07-20
Fixed newsletter signup on account creation page

cykolln:
https://klaviyo.tpondemand.com/entity/131521-m2-extension-not-syncing-newsletter-subscribers

Bug Description: when a user registered for an account on the account creation page the subscription wasn't properly being added to Klaviyo because `subscriber->hasDataChanges` was evaluating to false for new users. This change adjusts the logic to additionally check to see if the subscriber is new. We should now subscribe/unsubscribe registered users to Klaviyo lists when the subscription is either updated on the /customer/account/ page or if they checked the box on the /customer/account/create/ page.


tylerkemme (approved):
looks great!

PR #207, opened 2022-09-09
updating phpdoc params on ReclaimInterface method

wbendris-klaviyo:


cykolln:
Looks good! Would you mind updating the CHANGELOG and putting this fix in under the Unreleased heading? When we go to draft the release we can then just move it over to the new version.

wbendris-klaviyo:
> Looks good! Would you mind updating the CHANGELOG and putting this fix in under the Unreleased heading? When we go to draft the release we can then just move it over to the new version.

Sure thing, should be all set!

cykolln (approved):
Looks great!

PR #211, opened 2022-09-29
Added StoreName to custom profile property

olaf-dunn-klaviyo:
The changes for the PR are to add 'storeName' custom profile property during a Magento sync.
This will allow the customers to be able to create segments based upon their M2 Store Name.


cykolln:
Hey @olaf-dunn-klaviyo - Product some concerns about adding a profile property without any backfill. Ed requested that you make a ProductBoard request for this feature.

PR #214, opened 2022-10-25
Do not load not needed backend classes

ihor-sviziev:
## Description
This PR replaces the backend context with the frontend one. Since this controller is used on frontend, it doesn't make any sense to create classes needed for backend, e.g. backend session which is part of this context.

Basically, it just improves speed of this controller and reduces session locks on checkout 

## Manual Testing Steps

1.

## Pre-Submission Checklist:


invertedfjord (approved):

cykolln (changes requested):
Can you please update the changelog by adding this change under Unreleased - we can include it in the upcoming release.

cykolln (approved):

cykolln:
This has been included in the latest release 4.0.7, thank you for another contribution @ihor-sviziev - much appreciated!

PR #215, opened 2022-10-28
ATC bug fixes, updated cleanup crons, enforce payload size

cykolln:
## Description

This PR encompasses several improvements for the added to cart event. Namely it enforces the payload size, and doesn't allow for rows with truncated payload to be committed to the database. 

Previously, truncated Added to Cart payloads would raise unhandled exceptions when processing in both `Cron/EventsTopic.php` and `Cron/KlSyncs.php`, which in turn would prevent the processing of the remaining rows. The unhandled exceptions also prevented the rows that were able to move over to the kl_sync table from being marked as MOVED, leading to repetitive processing and continuous retries. This PR adds exception handling which will log the exception and mark any rows that raise exceptions during processing as FAILED. The cleanup crons have been adjusted to additionally delete FAILED rows from both the kl_events table and the kl_sync table. Previously these rows would remain indefinitely. 

These fixes are intended to prevent truncated payloads for future events and clear out failed events from the database. Future improvements will be made to further optimize the Added to Cart event. 

## Manual Testing Steps

1. Pre-change, recreated the issue with truncated payloads. Triggered a variety of Added to Cart events, some with truncated payloads, some without. 
2. After changes, verified that an event with a payload exceeding the maximum character count would not be added to the kl_events table and that a log was created. 
3. Verified that the `klaviyo_events_topic` cron processed all rows - marking those with existing truncated payloads as FAILED and marking all rows with valid payloads as MOVED.
4. Verified that the `klaviyo_webhook/event_sync` cron processed all rows - marking those with truncated payloads as failed. 
5. Verified that the 2 cleanup crons removed rows marked as FAILED for both tables, as well as MOVED for kl_events and SYNCED for kl_sync.

## Pre-Submission Checklist:

### DIFF
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -100,7 +100,7 @@ class EventsTopic
                 catch (\Exception $e) {
                     // the payload was likely truncated, this will catch any indexing errors that occur during processing
                     // defaults to a failed response and allows the other rows to continue syncing
-                    $this->_klaviyoLogger->log(sprintf("Unable to process Added to Cart data: %s", $e->getMessage()));
+                    $this->_klaviyoLogger->log(sprintf("[moveRowsToSync] Unable to process Added to Cart data: %s", $e->getMessage()));
                     array_push($idsFailed, $event['id']);
                     continue;
                 }
@@ -126,7 +126,7 @@ class EventsTopic
                 $sync->save();
                 array_push($idsMoved, $event['id']);
             } catch (\Exception $e) {
-                $this->_klaviyoLogger->log(sprintf("Unable to move row: %s", $e->getMessage()));
+                $this->_klaviyoLogger->log(sprintf("[moveRowsToSync] Unable to move row: %s", $e->getMessage()));
                 array_push($idsFailed, $event['id']);
             }
         }
--- a/Cron/EventsTopic.php
+++ b/Cron/EventsTopic.php
@@ -109,7 +109,7 @@ class EventsTopic
             if (strlen($event['payload']) > self::PAYLOAD_CHARACTER_LIMIT) {
                 // Above processing resulted in a payload size that exceeds the limit
                 // defaults to a failed response and allows the other rows to continue syncing
-                $this->_klaviyoLogger->log(sprintf("Dropped Event - payload too long, character count:%d",strlen($event['payload'])));
+                $this->_klaviyoLogger->log(sprintf("[moveRowsToSync] Dropping event - payload too long, character count: %d",strlen($event['payload'])));
                 array_push($idsFailed, $event['id']);
                 continue;
             }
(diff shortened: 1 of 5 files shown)


Cron/EventsTopic.php
```diff
@@ -85,11 +88,30 @@ public function moveRowsToSync()
...
         foreach ($eventsData as $event){
             if ($event['event'] == 'Added To Cart') {
-                $event['payload'] = json_encode($this->replaceQuoteIdAndCategoryIds($event['payload']));
+                try {
+                    $event['payload'] = json_encode($this->replaceQuoteIdAndCategoryIds($event['payload']));
+                }
+                catch (\Exception $e) {
+                    // the payload was likely truncated, this will catch any indexing errors that occur during processing
+                    // defaults to a failed response and allows the other rows to continue syncing
+                    $this->_klaviyoLogger->log(sprintf("Unable to process Added to Cart data: %s", $e->getMessage()));
```
smoucka, line 103: Can we make this logged error unique from the saving to the events table so we can identify where we're going over the column limit?

Cron/KlSyncs.php
```diff
@@ -124,29 +126,37 @@ private function sendUpdatesToApp(array $groupedRows, bool $isRetry = false)
...
-                    unset($decodedPayload['time']);
-
-                    //TODO: if conditional for backward compatibility, needs to be removed in future versions
-                    $storeId = '';
-                    if (isset($decodedPayload['StoreId']))
-                    {
-                        $storeId = $decodedPayload['StoreId'];
-                        unset($decodedPayload['StoreId']);
+                    try {
+                        $decodedPayload = json_decode($row['payload'], true);
```
smoucka, line 130: Can we check the return here? It shoudl be false or null if truncated and we cannot decode. Can log specifically for that.

Cron/KlSyncs.php
```diff
@@ -124,29 +126,37 @@ private function sendUpdatesToApp(array $groupedRows, bool $isRetry = false)
...
+                        }
+
+                        $response = $this->_dataHelper->klaviyoTrackEvent(
+                            $row['topic'],
+                            json_decode($row['user_properties'], true ),
+                            $decodedPayload,
+                            $eventTime,
+                            $storeId
+                        );
+                        if (!$response) {$response = '0';}
```
smoucka, line 150: Can we add a comment here?

Cron/KlSyncs.php
smoucka: If we know we've got a bad payload can we skip sending to Klaviyo in the first place?

Cron/KlSyncs.php
```diff
@@ -124,29 +126,37 @@ private function sendUpdatesToApp(array $groupedRows, bool $isRetry = false)
...
+
+                        array_push($responseManifest["$response"], $row['id']);
+                    } catch ( \Exception $e) {
+                        // the payload was likely truncated, this will catch any indexing errors
+                        // defaults to a failed response and allows the other rows to continue syncing
```
smoucka, line 155: Can we update this to reflect why this might fail if we're not trying to send truncated payloads in the first place?

smoucka (approved):

PR #222, opened 2023-01-09
TP157110: Add requirement for lowest supported php version

award28:
## Description

Adds our currently lowest supported version as a requirement in our composer.json file.

## Manual Testing Steps

1.

## Pre-Submission Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
         "optimize-autoloader": true
     },
     "require": {
+        "php": ">=7.1.0",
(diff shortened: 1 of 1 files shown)


chavesj (approved):

cykolln (commented):
given version 4.0.0+ of our extension is only compatible with magento versions 2.4.0+ we should go off of the minimum php version requirement for that m2 version which would be 7.3. However, I don't know if it's best practice to specify php version in extension composer.json files - let me look into this.

PR #226, opened 2023-01-13
Subscriber source

natewiebe13:
## Description
This adds Magento 2 as the source when customers subscribe to the newsletter.

Logging $source properties on Klaviyo profiles is necessary to track where contacts are originating from and understanding their various user journeys, similar to understanding traffic source patterns in Google Analytics.

Source ID used was supplied by Klaviyo Support and can be referenced here: https://help.klaviyo.com/hc/en-us/articles/1260804673530-Source-Value-Reference

## Manual Testing Steps

1. Subscribe to the newsletter.
2. The subscriber's profile should have Magento 2 specified as the source

## Pre-Submission Checklist:


natewiebe13:
@cykolln any chance we can get this in for the next release? It's required for CASL compliance and we currently have to maintain patches for client projects.

chavesj (approved):
LGTM, sorry for the delay

PR #227, opened 2023-01-24
TP156258: Switch from _learnq to klaviyo object in javascript

award28:
## Description

Updated the legacy `_learnq` js object to the new `klaviyo` js object.

## Manual Testing Steps

1. Validated the `klaviyo` object is available when the klaviyo.js script is installed.

## Pre-Submission Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### [Unreleased]
 
+#### Changed
+- Updated the legacy `_learnq` js object to the new `klaviyo` js object.
+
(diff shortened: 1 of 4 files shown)


award28:
I'm not going to create a release just for this change - it can go out with the next scheduled release.

smoucka (approved):

toddbc:
`klaviyo.isIdentified()` returns a promise so I think this introduces some issues.  `_learnq.isIdentified()` did not return a promise.

cykolln:
> `klaviyo.isIdentified()` returns a promise so I think this introduces some issues. `_learnq.isIdentified()` did not return a promise.

@toddbc thank you for raising this - we are rolling this change back with a plan to fix and include it in a future release

PR #228, opened 2023-02-08
TP159713: Create pre-commit config and run on all files

award28:
## Description

Add the pre-commit config and execute over all files to conform to the PSR-12 standard.

## Manual Testing Steps
Ran pre-commit over all files.
Screen Shot 2023-02-09 at 1 13 20 PM

## Pre-Submission Checklist:

  - [ ] Increment the version in the following places: module.xml and composer.json


composer.json
```diff
@@ -15,7 +15,18 @@
         "optimize-autoloader": true
     },
     "require": {
-        "magento/module-quote": ">=101.1.3",
```
smoucka, line 18: Curious why we're removing this dependency?

award28: That's a great question! I'm meeting with Claire to discuss if we can. When I searched online, I couldn't find a reference to this plugin in packagist, magento slack community, or POGS (_plain old google search_). It looks like it was added two years ago, and it may just in included as part of magento 2's source code now.

award28: To update this for posterity: We decided to add a `composer.dev.json` for local development, which will get used within the makefile as `make composer-dev`. More [here](https://klaviyo.slack.com/archives/C046U293FDM/p1675960628020699). and in PR #230.

smoucka (approved):
🔥

PR #231, opened 2023-02-15
202302 tp165376

cykolln:
## Description

This PR fixes a bug where customers that hadn't confirmed their email for their Site account were getting unsubscribed in Klaviyo. This was happening because we weren't accounting for the other two order statuses in our observer file - as a result every subscription update where the customer wasn't explicitly subscribed resulted in them getting unsubscribed. 

## Manual Testing Steps

Tested the following scenerios: 
1. Create a new account for a site that has account confirmation required enabled and opt in to the newsletter. The Klaviyo newsletter settings are set to use the Klaviyo Double Opt-in. 
- Verify that the profile has not been created on Klaviyo yet
- Confirm account via the email
- Verify that I've received the Klaviyo DOI email and once that is confirmed that the profile has been added to the list
2. Create a new account for a site that does **not** have account confirmation required enabled and opt in to the newsletter. The Klaviyo newsletter settings are set to use the Klaviyo Double Opt-in. 
- Verify that I've received the Klaviyo DOI email and once that is confirmed that the profile has been added to the list correctly in Klaviyo. 
3. From an existing account on the store, update the newsletter settings to opt-out of emails.
- verify that the profile has been unsubscribed in Klaviyo as expected.
4. From an existing account on the store, update the newsletter settings to opt in to emails.
- verify that the profile has been resubscribed to klaviyo as expected after confirming via the DOI email. 
5. While not logged into an account on the magento store, fill out the footer form
- Verify that I've received the Klaviyo DOI email and once that is confirmed that the profile has been added to the list correctly in Klaviyo. 
- 
## Pre-Submission Checklist:

### DIFF
--- a/Observer/NewsletterSubscribeObserver.php
+++ b/Observer/NewsletterSubscribeObserver.php
@@ -47,17 +47,20 @@ class NewsletterSubscribeObserver implements ObserverInterface
 
         /** @var Subscriber $subscriber */
         $subscriber = $observer->getDataObject();
-        if ($subscriber && ($subscriber->hasDataChanges() || $subscriber->isObjectNew())) {
+        if ($subscriber && $subscriber->isStatusChanged()) {
+            $subscriptionStatus = $subscriber->getStatus();
             $customer = $this->getCustomer($subscriber);
 
-            if ($subscriber->isSubscribed()) {
+            if ($subscriber->getId() && $subscriptionStatus === Subscriber::STATUS_SUBSCRIBED) {
                 $this->helper->subscribeEmailToKlaviyoList(
                     $customer ? $customer->getEmail() : $subscriber->getEmail(),
                     $customer ? $customer->getFirstname() : $subscriber->getFirstname(),
                     $customer ? $customer->getLastname() : $subscriber->getLastname(),
                     self::SOURCE_ID_MAGENTO2
                 );
-            } else {
+            }
+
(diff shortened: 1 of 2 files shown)


Observer/NewsletterSubscribeObserver.php
```diff
@@ -45,16 +45,19 @@ public function execute(Observer $observer)
 
         /** @var Subscriber $subscriber */
         $subscriber = $observer->getDataObject();
-        if ($subscriber && ($subscriber->hasDataChanges() || $subscriber->isObjectNew())) {
+        if ($subscriber && $subscriber->isStatusChanged()) {
+            $subscriptionStatus = $subscriber->getStatus();
             $customer = $this->getCustomer($subscriber);
 
-            if ($subscriber->isSubscribed()) {
+            if ($subscriber->getId() && $subscriptionStatus === Subscriber::STATUS_SUBSCRIBED) {
```
smoucka, line 52: Just curious why we need to check the id?

cykolln: I was going off of the `isSubscribed` function from the core newsletter module: https://github.com/magento/magento2/blob/2.4-develop/app/code/Magento/Newsletter/Model/Subscriber.php#L377. I removed that function call in favor of explicit status checks for subscribed and unsubscribed

smoucka: Awesome, sounds good!

smoucka (approved):

Issue #191, opened 2022-05-24
Klaviyo blocks checkout from working - error with method getKlSmsConsent()


RoboSowo:
We are in the process of installing Klaviyo to our Magento2, but before even configuring the app, we can see that the app is preventing our checkout flow in functioning. We get following error message: 

main.CRITICAL: Error: Call to undefined method Magento\Checkout\Api\Data\ShippingInformationExtension::getKlSmsConsent() in /home/web/OURSITE/releases/20220523065500/vendor/klaviyo/magento2-extension/model/Checkout/ShippingInformationManagement.php:29

Does anyone experience something similar and do you have a fix for it?

We previously installed Klaviyo to our test site, and that worked as expected. We have done exactly same steps so far.

Br Sophie

VivekShingala:
Hi @RoboSowo 

We are also experiencing the same issue. Our staging site is working fine, but it is breaking the production site checkout page with the error you have mentioned.

We have same Magento, PHP and Klaviyo versions for our staging and production environment. Could you please help me if you have already fixed this issue?

RoboSowo:
No, we did not find a fix, so we had to install an older version of the
application...

Do you plan to make a solution in this in the near future?

Br Sophie

VivekShingala:
Oh, I see.

Yes, I am working on it to find and fix the issue. Will let you know here.

cykolln:
Thank you for raising this - I've created an internal ticket to investigate this and will update this issue when it gets picked up.

cykolln:
@RoboSowo @VivekShingala can you please share which version of Magento you're using?

VivekShingala:
Hi @cykolln 

I'm using below versions for my website:

Magento: 2.4.3-p2
PHP: 7.4.12
Klaviyo: 4.0.5

Please note that we have the same configuration for our staging site as well and it's working correctly. The issue we are facing is on production server.

cykolln:
Hi @VivekShingala - thanks for supplying your configuration for us. A lot of the time issues with extension attributes can be resolved by removing the /generated directory and recompiling:

```
rm -rf generated/*
php bin/magento setup:di:compile
```

VivekShingala:
Hi @cykolln 

We are using deployer so generated folder is newly created each time we deploy the code to the server. So, that can't be the issue.

wmdkdkussin:
FYI: We still experience the issue with KLYVIYO v4.0.11 and Magento v2.4.2-p2.

[...]

PR #244, opened 2023-03-30
Fix shipping information update on checkout for klEmailConsent and klSmsConsent fields

dadolun95:
## Description
PR preventing Klaviyo to block checkout from working - error with method getKlSmsConsent() and getklEmailConsent() already issued on #191 

## Environment
Magento 2.4.5
PHP 8.1

## Manual Testing Steps
Enable the module, then go at "Stores > Configuration > Klaviyo > Consent at Checkout" section and flag "Subscribe contacts to email marketing at checkout" to "Yes" or/and  flag "Subscribe contacts to SMS marketing at checkout" to "Yes". 
Save the configurations, clean config cache.
After that go on checkout as guest and try complete an order.

## Expected Results
Before this fix, magento checkout will block you sending a 500 error on rest/default/V1/guest-carts/shipping-information endpoint and showing the following message (if your magento is in developer mode, webapi report error on production instead):
<< Property "KlEmailConsent" does not have accessor method "getKlEmailConsent" in class "Magento\Quote\Api\Data\AddressExtensionInterface". >> or << Property "KlSmsConsent" does not have accessor method "getKlSmsConsent" in class "Magento\Quote\Api\Data\AddressExtensionInterface". >> depending on "Consent at Checkout" configuration.
After the fix with same procedure you'll be able to complete the order normally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- BEGIN RELEASE NOTES -->
 ### [Unreleased]
 
+#### Fixed
+- Fixed bug in checkout where shipping information update caused report error if email or sms consent where activated
+
 #### Changed
 - Fixed bug where historical sync wouldn't run for index of 0
 
--- a/etc/extension_attributes.xml
+++ b/etc/extension_attributes.xml
@@ -4,6 +4,10 @@
         <attribute code="kl_sms_consent" type="string"/>
         <attribute code="kl_email_consent" type="string"/>
     </extension_attributes>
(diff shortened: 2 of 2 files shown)


cykolln (changes requested):
Thank you for contributing! Can you please update the changelog and add this update to the Unreleased section - we will include this in the release we have planned for next week. You can use the changelogged tool mentioned in the README by running `changelogger unreleased add`.

dadolun95:
Hi @cykolln 
CHANGELOG file updated as you requested. 
Hope to see my fix on the official repository next week :+1: 
Wish you a very nice weekend!

cykolln (approved):

PR #254, opened 2023-04-26
Improve performance of Viewed Product block

ihor-sviziev:
## Description

I noticed that some product pages loaded pretty fast while others did not. I did profiling and noticed that Klaviyo Viewed Product block was loading about 0.26-0.3 seconds on my pretty fast machine, which is very big number. That was caused by loading all linked category models in the loop (some products have 150+ categories).
To improve this situation, I replaced the loading category model in the loop with using a single collection call with fetching only needed info (category name). That worked way faster on the same machine- about 0.007 vs 0.3 seconds.

## Manual Testing Steps

1. Create product and link it to 150+ categories
2. Enable Magento profiler with `php bin/magento dev:profiler:enable` and see how long klaviyo block is getting rendered.
3. Apply the patch
4.  Compare results between the measurements 
 
## Pre-Submission Checklist:

### DIFF
--- a/Block/Catalog/Product/ViewedProduct.php
+++ b/Block/Catalog/Product/ViewedProduct.php
@@ -4,7 +4,6 @@ namespace Klaviyo\Reclaim\Block\Catalog\Product;
 
 use Klaviyo\Reclaim\Helper\ScopeSetting;
 use Magento\Catalog\Helper\Image;
-use Magento\Catalog\Model\CategoryFactory;
 use Magento\Framework\Registry;
 use Magento\Framework\View\Element\Template;
 use Magento\Framework\View\Element\Template\Context;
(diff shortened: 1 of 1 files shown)


ihor-sviziev:
I see a related PR https://github.com/klaviyo/magento2-klaviyo/pull/130, but it has a limit of 30 categories. I am wondering if that's expected. In comparison, my PR isn't adding any difference to the result.

iGerchak (approved):

cykolln (commented):
Thank you for contributing! Happy to approve, can you please update the changelog and add this update to Unreleased under Fixed. You can use the changelogger tool: `changelogger unreleased add` to add this change

cykolln (approved):

PR #263, opened 2023-09-18
4.0.12 - v3 upgrade

jordanleslie:
## Description

Introduced a new class, `KlaviyoV3Api` dedicated to handling API requests and responses for Klaviyo V3. This class was modeled after @siddwarkhedkar [Prestashop V3 PR.](https://github.com/klaviyo/prestashop_klaviyo/pull/124)
This class is responsible for transforming request data to match the structure of the Klaviyo V3 API.
* Updated API Endpoint URLs
* Updated and centralized the URL endpoints for the Klaviyo V3 API within the KlaviyoV3Api class for consistency.
* Added several new exception types tailored to specific error scenarios, improving the precision of error reporting and handling.
* New Method handleApiResponse:

* Introduced a new method called handleApiResponse in the KlaviyoV3Api class.
This method consistently handles API responses, parsing JSON data, and throwing appropriate exceptions when necessary.

## Manual Testing Steps

1. Tested on version 2.3.7, 4.0.12

## Pre-Submission Checklist:

  - [x] The links in the changelog have been updated to point towards the new versions


siddwarkhedkar (changes requested):

KlaviyoV3Sdk/KlaviyoV3Api.php
```diff
@@ -0,0 +1,490 @@
...
+    public function addProfileToList($list_id, $profile_id)
+    {
+        $body = array(
+            self::DATA_KEY_PAYLOAD => array(
+                self::TYPE_KEY_PAYLOAD => self::PROFILE_KEY_PAYLOAD,
+                self::ID_KEY_PAYLOAD => $profile_id
+            )
+        );
+
+        return $this->requestV3("api/lists/$list_id/relationships/profiles/", self::HTTP_POST, $body);
```
ali-aslam, line 186: Need error handling for when this V3 API call fails. Or if the purpose here is to just send back the $response_body from the API call then the code that's making the addProfileToList() method call needs to handle the condition where the response is unexpected.

ali-aslam: Same comment for all the other places you are making the V3 API calls. We want to make sure the failure/unexpected cases are handled.

ali-aslam (changes requested):

KlaviyoV3Sdk/KlaviyoV3Api.php
```diff
@@ -0,0 +1,490 @@
...
+     * https://developers.klaviyo.com/en/v2023-08-15/reference/get_lists
+     *
+     * @return array
+     * @throws KlaviyoApiException
+     * @throws KlaviyoAuthenticationException
+     * @throws KlaviyoRateLimitException
+     */
+    public function getLists()
+    {
+        $response_body = $this->requestV3('api/lists/', self::HTTP_GET);
```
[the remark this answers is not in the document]
jordanleslie: Thank you! I took a deeper look at error cases for this method. V3 Api gives us a consistent error object which we're handling within [requestV3](https://github.com/klaviyo/magento2-klaviyo/blob/a6cb08a790a8d3e9770cc219c0eb6d333c5e42a7/KlaviyoV3Sdk/KlaviyoV3Api.php#L330). In the event that we get an empty list array were still able to pass that forward to the front end.

KlaviyoV3Sdk/KlaviyoV3Api.php
siddwarkhedkar: We don't need the response from this method. This should only make the call to add the profile

KlaviyoV3Sdk/KlaviyoV3Api.php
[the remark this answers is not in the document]
siddwarkhedkar: If the request responds with anything but a 200 - 300 that is covered by the requestV3 method. For other scenarios, when there are no lists returned by this endpoint but with a 200 status code (no lists exists in Klaviyo), this is handled by the caller already. 

No action required here

KlaviyoV3Sdk/KlaviyoV3Api.php
```diff
@@ -341,20 +341,18 @@ protected function requestV3($path, $method = null, $body = null, $attempt = 0)
         }
 
         $response = curl_exec($curl);
-
+        $phpVersionHttpCode = version_compare(phpversion(), '5.5.0', '>') ? CURLINFO_RESPONSE_CODE : CURLINFO_HTTP_CODE;
+        $statusCode = curl_getinfo($curl, $phpVersionHttpCode);
         // In the event that the curl_exec fails for whatever reason, it responds with `false`,
         // Implementing a timeout and retry mechanism which will attempt the API call 3 times at 5 second intervals
-        if (!$response) {
+        if ($statusCode != 202) {
```
siddwarkhedkar, line 348: statusCode 
```suggestion
        if ($statusCode < 200 or $statusCode > 300) {
```

siddwarkhedkar (approved):

ali-aslam (approved):

cykolln (approved):

PR #275, opened 2023-11-02
Add name to checkout analytics block

norgeindian:
## Description

With this adaption, it is possible to reference the block, so it can be adapted to custom needs.

### DIFF
--- a/view/frontend/layout/checkout_index_index.xml
+++ b/view/frontend/layout/checkout_index_index.xml
@@ -2,7 +2,7 @@
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <body>
         <referenceBlock name="content">
-            <block class="Klaviyo\Reclaim\Block\Initialize" template="Klaviyo_Reclaim::analytics/initialize.phtml" name="klaviyo.checkout.initialize" />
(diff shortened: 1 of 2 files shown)


cykolln (changes requested):
@norgeindian thanks for contributing! Could you please add a line to the changelog under Unreleased describing this change? You can use our [changelogger](https://pypi.org/project/changelogged/) tool and run `changelogger add`. Then we can get this released with our next extension version.

norgeindian:
@cykolln, sure, is it okay like this?

view/frontend/layout/checkout_index_index.xml
```diff
@@ -2,7 +2,7 @@
 <page xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="urn:magento:framework:View/Layout/etc/page_configuration.xsd">
     <body>
         <referenceBlock name="content">
-            <block class="Klaviyo\Reclaim\Block\Initialize" template="Klaviyo_Reclaim::analytics/initialize.phtml" />
+            <block class="Klaviyo\Reclaim\Block\Initialize" template="Klaviyo_Reclaim::analytics/initialize.phtml" name="klaviyo.checkout.initialize" />
```
cykolln, line 5: sorry for the delay on this review - can you update this to `klaviyo_reclaim.checkout.initialize` just to follow our pattern for other block names

norgeindian: @cykolln , of course. Okay like that?

cykolln (approved):

cykolln:
@norgeindian Thanks so much, we will get this out in the next release!

PR #279, opened 2023-12-11
Paginate lists for newsletter setup

smoucka:
## Description

V3 lists api only returns 10 lists per page. In order to fully populate the Newsletter configuration dropdown menu we need to paginate to fetch all lists for an account.

## Manual Testing Steps

1. local setup, tested with an account with 155 lists and with an account with 4 lists. 155 had a few seconds of latency due to the synchronous requests

## Pre-Submission Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,7 +187,7 @@ fabric.properties
 
 .idea/*
 
-!.idea/codeStyles
+#!.idea/codeStyles
 !.idea/runConfigurations
 
 ### PHPUnit ###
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -187,7 +187,7 @@ fabric.properties
 
 .idea/*
 
-!.idea/codeStyles
+#!.idea/codeStyles
```
jordanleslie, line 190: Are we removing this from the .gitignore?

smoucka: Yea, I'm not sure why that was excluded. The `.idea` directory is phpstorm related. I ended up with a codeStyles config file that was committed so making sure we don't get those in here by accident.

cykolln (approved):

PR #286, opened 2024-01-30
Bug fixes for v3 api release

cykolln:
## Description

This PR addresses several bugs from 4.1.0 release which added support for Klaviyo's V3 APIs. 

* Fixes issues with store-scoped requests. If you scope each store to a different Klaviyo account we will now correctly retrieve the correct private API key and sync the events. 
* Fixes response handling for Added to Cart requests - previously our APIs returned 0/1 for failure/success. Now we return json with more information about the errors (if applicable). This updates our handling to accept the new response schema and correctly marks successful rows as such. 
* Fixes logging to log out a much more useful error message on retries, adds handling for 401s
* Updates Added to Cart observers to check for the private api key instead of the public key. 
* Fixes a bug w/ directly adding a profile to a list

## Manual Testing Steps

- Tested the following cases: 
     1. Default scoping, multiple sites: ensured all sites correctly fired ATC events using the default config. 
     2. Store-scoped, multiple sites, multiple Klaviyo accounts: ensured all sites correctly fired ATC events using the store-scoped config. Verified events appeared on correct Klaviyo account
     3. Multiple sites, one using default, one using store-scoped config: ensured all sites correctly fired ATC events using the store-scoped config. Verified events appeared on correct Klaviyo account

- Tested error handling cases (forced 401s, 403s) to validate RETRY was set as row status and that logs were emitted with the correct message. 
- Tested missing Private Key, validated no row was added to the kl_events table. 

## Pre-Submission Checklist:

- [x] **Internal Only** - If this is a release, please confirm the following:

### DIFF
--- a/Observer/SalesQuoteSaveAfter.php
+++ b/Observer/SalesQuoteSaveAfter.php
@@ -76,10 +76,9 @@ class SalesQuoteSaveAfter implements ObserverInterface
         $kl_decoded_cookie = json_decode(base64_decode($_COOKIE['__kla_id']), true);
 
         // Get the custom variable set in the DataHelper object via the SalesQuoteProductAddAfter observer.
-        // Check if the public key and Added to Cart payload are set
-        $public_key = $this->_scopeSetting->getPublicApiKey();
+        // Check if the private key and Added to Cart payload are set
         $klAddedToCartPayload = $this->_dataHelper->getObserverAtcPayload();
-        if (!isset($klAddedToCartPayload) or !isset($public_key)) {
+        if (!isset($klAddedToCartPayload) or empty($this->_scopeSetting->getPrivateApiKey())) {
             return;
         }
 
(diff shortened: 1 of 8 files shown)


Observer/SalesQuoteSaveAfter.php
```diff
@@ -76,10 +76,9 @@ public function execute(Observer $observer)
         $kl_decoded_cookie = json_decode(base64_decode($_COOKIE['__kla_id']), true);
 
         // Get the custom variable set in the DataHelper object via the SalesQuoteProductAddAfter observer.
-        // Check if the public key and Added to Cart payload are set
-        $public_key = $this->_scopeSetting->getPublicApiKey();
+        // Check if the private key and Added to Cart payload are set
         $klAddedToCartPayload = $this->_dataHelper->getObserverAtcPayload();
-        if (!isset($klAddedToCartPayload) or !isset($public_key)) {
+        if (!isset($klAddedToCartPayload) or empty($this->_scopeSetting->getPrivateApiKey())) {
```
jordanleslie, line 81: This call required the private api key rather than the public key?

jordanleslie (approved):
Awesome, thank you! One question but otherwise LGTM

Observer/SalesQuoteSaveAfter.php
cykolln: yes! we are calling the /events API: https://developers.klaviyo.com/en/reference/create_event

PR #291, opened 2024-02-06
adds fix for custom props on track requests

cykolln:
## Description

Fixes issue with custom properties. Although `track` isn't called from the klaviyo module code, we are including it in the patch release so folks that have customized their m2 module version to include custom track requests can easily migrate these over to the new APIs. 

## Manual Testing Steps

1.

## Pre-Submission Checklist:


klaviyoit:
https://klaviyo.tpondemand.com/entity/3

jordanleslie (approved):

PR #298, opened 2024-05-21
Updates Added to Cart event to use simple product images where applicable

cykolln:
## Description

Previously we synced the configurable product image path instead of the simple product that was added to the cart. This PR updates that behavior to use the simple product image if applicable, and fall back on the configurable product if we cannot find an image on the simple product. 

It also adds 2 new event fields for Added to cart: StoreId and AddedSimpleProductId. These are stored at the top level and can be used for additional filtration or use in catalog tags in email templates. 

## Manual Testing Steps
1. Tested simple products, bundles, groups, and configurable products. Made sure the correct fields were syncing to klaviyo. Validated that the behavior for bundles and grouped products was unchanged. Verified that simple product information is syncing correctly for simple product and configurable product carts. 

1. Tested 

## Pre-Submission Checklist:

### DIFF
--- a/Cron/KlSyncs.php
+++ b/Cron/KlSyncs.php
@@ -146,7 +146,6 @@ class KlSyncs
                         $storeId = '';
                         if (isset($decodedPayload['StoreId'])) {
                             $storeId = $decodedPayload['StoreId'];
-                            unset($decodedPayload['StoreId']);
                         }
 
                         $response = $this->_dataHelper->klaviyoTrackEvent(
--- a/Observer/SalesQuoteProductAddAfter.php
+++ b/Observer/SalesQuoteProductAddAfter.php
@@ -201,7 +201,8 @@ class SalesQuoteProductAddAfter implements ObserverInterface
      * @param $addedItem
      * @return mixed
      */
-    public function getSimpleProductForEvent($addedItem) {
+    public function getSimpleProductForEvent($addedItem)
+    {
         # try to grab the simple product
         $simpleProduct = null;
         try {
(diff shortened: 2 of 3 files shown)


Cron/KlSyncs.php
```diff
@@ -146,7 +146,6 @@ private function sendUpdatesToApp(array $groupedRows, bool $isRetry = false)
                         $storeId = '';
                         if (isset($decodedPayload['StoreId'])) {
                             $storeId = $decodedPayload['StoreId'];
-                            unset($decodedPayload['StoreId']);
```
cykolln, line 149: Adding this to the payload because we may want to use it down the line for variant triggered flows in the app

Observer/SalesQuoteProductAddAfter.php
```diff
@@ -111,13 +125,17 @@ public function klBuildCartData($quote, $addedItem)
...
+
             $cartItemId = $product->getId();
             $itemCategories = $product->getCategoryIds();
             $itemName = $item->getName();
             $currentProduct = [
                 'Categories' => (array) $itemCategories,
-                'ImageUrlKey' => (string) is_null($product->getData('small_image')) ? "" : stripslashes($product->getData('small_image')),
+                'ImageUrlKey' => $this->getImagePreferringVariant($product, $simpleProduct),
                 'ProductId' => (int) $cartItemId,
+                'SimpleProductId' => (int) is_null($simpleProduct) ? null : $simpleProduct->getId(),
```
smoucka, line 138: Just confirming if this is doing what we want. If `$simpleProduct` is null are we hoping to return `null` or `0` here? I don't believe we need to cast the conditional expression and that appears to be what's happening here.

smoucka (approved):
One nit but otherwise lgtm!

Issue #295, opened 2024-04-15
E-Mail not encoded in searchProfileByEmail


spicyavocadooo:
### Environment details

PHP version: 8.1
Magento version: 2.4.6-p4
Klaviyo extension version: 4.1.3
Anything else that would help a developer reproduce the bug?

### Steps to reproduce

Create customer with email address including "+" symbol
Unsubscribe Customer
Subscribe Customer again

### Expected result

Customer is subscribed to newsletter after subscribing again

### Actual result

Customer is not subscribed to newsletter after subscribing again, exception in log 
"Unable to subscribe ... to list ...: Klaviyo\Reclaim\KlaviyoV3Sdk\Exception\KlaviyoApiException: A profile already exists with one of these identifiers."

### Additional information

searchProfileByEmail is using normal email and not encoded email as necessary for the API

cykolln:
Thanks for opening - we need to add `urlencode` around the email in `searchProfileByEmail`. We've added this to our backlog and can include it in our next release

cykolln:
This has been fixed in 4.1.4

PR #299, opened 2024-05-22
Adds urlencode to email before search request

cykolln:
## Description

Fixes a bug where we weren't url encoding emails before requesting the `api/profiles/` endpoint. For emails containing special characters this was resulting in misses on the search and the following request to create the new profile would fail because it does already exist. 

## Manual Testing Steps

1. Selected to **not** use Klaviyo's opt-in settings in the extension settings. Then filled out the M2 footer form with an email that already existed in Klaviyo, made sure it was properly added to list. 
2. Tested the same as above but with a new email that didn't exist in Klaviyo. 

## Pre-Submission Checklist:


cykolln:
Fixes https://github.com/klaviyo/magento2-klaviyo/issues/295

kamidzi-klaviyo-integrations (approved):
lgtm.

PR #331, opened 2025-05-21
Enhance subscription methods to accept store ID for better context handling

DianaSol:
## Description

Store-Aware Newsletter Sync:

- The methods subscribeEmailToKlaviyoList and unsubscribeEmailFromKlaviyoList in Data.php now accept an optional $storeId parameter.

- The $storeId is used to fetch the correct Klaviyo list and opt-in settings, ensuring that newsletter subscriptions and unsubscriptions are performed in the context of the correct Magento store.

## Manual Testing Steps

Create different lists for differences stores. Place an order from any store and check that subscriber goes into the appropriate list.

### DIFF
--- a/Helper/Data.php
+++ b/Helper/Data.php
@@ -93,12 +93,13 @@ class Data extends \Magento\Framework\App\Helper\AbstractHelper
      * @param string|null $firstName
      * @param string|null $lastName
      * @param string|null $source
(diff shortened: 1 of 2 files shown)


kamidzi-klaviyo-integrations:
@DianaSol  - could you please add some tests?

PR #337, opened 2025-07-24
Add isKlaviyoEnabled to product/viewed.phtml

GJHypostore:
## Description
Add isKlaviyoEnabled to product/viewed.phtml. The logic in this phtml file gets executed even if you disable the module. Other phtml files do have an check.

Manual Testing Steps
Verified that product/viewed.phtml executes regardless of whether the Klaviyo module is enabled or disabled.

Added a check for isKlaviyoEnabled at the top of product/viewed.phtml, matching the logic used in other phtml files.

Confirmed that the Klaviyo script no longer renders when the module is disabled.

Confirmed that the script still renders correctly when the module is enabled.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -34,6 +34,8 @@ jobs:
       - uses: actions/setup-python@v4.4.0
         with:
           python-version: '3.11'
+      - name: Install click 8.2.2 until bug in 8.3.0 is resolved
+        run: pip install click==8.2.2
       - name: Install script dependencies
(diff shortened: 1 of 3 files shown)


smoucka (approved):
Thank you @GerritJanHypoStore . Can you please update CHANGELOG.md as well?

GerritJanHypoStore:
@smoucka I've updated the CHANGELOG.md and version numbers. Please let me know if there's anything else you'd like me to adjust.

GJHypostore:
@smoucka No worries at all, I’ve gone ahead and rebased as requested.

PR #365, opened 2026-04-24
E2E: wait for newsletter POST response instead of flash message

smoucka:
## Summary

`fillOutNewsletterFooterForm` in the E2E helpers waited on `.message-success` after submitting the storefront footer subscribe form. With the "Honor Klaviyo consent" admin setting enabled, the module's subscribe path doesn't reliably render a Magento flash message on the post-redirect homepage — the POST returns 302, the homepage GET returns 200, but the messages KO container stays empty (likely FPC / customer-data `messages` section not refreshing for this code path in the local bitnami env). Result: the test timed out even though the Klaviyo profile was subscribed correctly.

Switch the helper to wait on the `/newsletter/subscriber/new` POST response. The downstream `checkProfileInKlaviyo` backoff remains the real source of truth for subscription success; the POST waiter is just there to fail loudly if the form submit never reaches the server.

## Test plan

- [ ] `npx playwright test e2e/api/subscribe-profile.spec.js` passes locally
- [ ] "Honor Klaviyo consent = Yes" and "= No" suites both still assert the correct Klaviyo consent state
- [ ] Confirm no regression in other specs that use `Storefront` (e.g. add-to-cart, account creation) — `fillOutNewsletterFooterForm` is only called from `subscribe-profile.spec.js`

## Follow-ups (not in this PR)

- Two pre-existing bugs in the module surfaced during diagnosis (retry-without-`return` in `KlaviyoV3Sdk/KlaviyoV3Api.php:381`, silent exception swallow in `Helper/Data.php:186`). Worth a separate ticket; noted locally in `NOTES_MODULE_BUGS.md`.
- Admin-menu flakiness when `setupAdminConfig` clicks "Stores → Configuration" while already on the Configuration page. Fix would be to `page.goto` the config section URL directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Test/e2e/locators/storefront.js
+++ b/Test/e2e/locators/storefront.js
@@ -57,14 +57,22 @@ class Storefront {
     }
 
     async fillOutNewsletterFooterForm() {
-        // Fill in the newsletter subscription form in the footer
         await this.page.fill('#newsletter', this.email);
 
-        // Click the subscribe button
+        // Wait for the subscribe POST rather than a Magento flash message —
+        // when "Honor Klaviyo consent" is enabled the module's code path
+        // doesn't reliably render the `.messages` flash on the redirect
+        // target. The Klaviyo API assertion downstream is the real check.
+        const postResponsePromise = this.page.waitForResponse(
+            resp => resp.url().includes('/newsletter/subscriber/new') &&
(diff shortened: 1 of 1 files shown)


zchenk (approved):
lgtm

